PR #1, opened 2014-12-04
make vmag^2 return NonNegReal and add type for vmag

AlexKnauth:


samth:
Was this made obsolete by the plot rewrite?

ntoronto:
Yes.

PR #5, opened 2015-04-17
Separate legend font from plot font for more control. 

deeglaze:
Unset parameters default to plot parameters.

Locally tested to make plots for my dissertation.


rfindler:
This looks reasonable to me. (Probably someone with more knowledge about plot should also confirm, however.)

alex-hhh:
Remaining work on this pull request:

* [ ] resolve conflicts with master
* [ ] the new plot parameters must be added to one of the parameter groups, otherwise the legend will just revert to using the default font parameters when `plot-snip` is used.  See https://github.com/racket/plot/blob/5d880bd4890911af0ff59c449b74005b6971fafb/plot-lib/plot/private/common/parameter-groups.rkt#L34, 
* [ ] add a test case  illustrating the use of the new parameters.  For an example, see https://github.com/racket/plot/tree/master/plot-test/plot/tests/PRs/46.rkt

Given that this is an old pull request, I will work on these tasks, as time allows.

alex-hhh:
Here is a sample use case:

```racket
#lang racket
(require plot)
(plot-width 300)
(plot-height 300)
(parameterize ([plot-font-size 8]
               [plot-font-family 'default]
               [plot-legend-font-size 14]
               [plot-legend-font-family 'modern])
  (plot (function sin -5 5 #:label "sin(x)")))
```

pr5

PR #10, opened 2016-01-03
Jitter options & parameters (for plotting points)

bennn:
New optional arguments for `points` and `points3d`, and new parameters to match.

When `> 0`, parameters `x-jitter`, `y-jitter`, and `z-jitter` will add random noise to each point along the specified dimension.
New points lie within a rectangle with width `x-jitter`, height `y-jitter`, and depth `z-jitter` centered at each point's original location, but kept within the bounds for x-min, x-max, etc. given to `points` or `points3d`.

Inspired by R's [ggplot2](http://zevross.com/blog/2014/05/05/unhide-hidden-data-using-jitter-in-the-r-package-ggplot2/). I think this is a nice substitute for boxplots and violin plots.

Example:
```
#lang racket
(require plot)

(plot
 (points (for/list ([_i (in-range 999)])
           (list (* 10 (random)) 2))
         #:alpha 0.4
         #:y-jitter 0.5
         #:sym 'fullcircle1
         #:color "red")
 #:x-min 0 #:x-max 10 #:y-min -5 #:y-max 5)
```

screen shot 2016-01-03 at 1 28 01 am

### DIFF
--- a/plot-doc/plot/scribblings/renderer3d.scrbl
+++ b/plot-doc/plot/scribblings/renderer3d.scrbl
@@ -56,8 +56,11 @@ For example, a scatter plot of points sampled uniformly from the surface of a sp
                             #:altitude 25)]
 
 When @(racket x-jitter), @(racket y-jitter), or @(racket z-jitter) is non-zero,
- each point @(racket p) is translated to a random location inside a rectangle centered at @(racket p) with width @(racket x-jitter), height @(racket y-jitter), and depth @(racket z-jitter).
+ each point @(racket p) is translated to a random location inside a box centered at @(racket p) with width @(racket x-jitter), height @(racket y-jitter), and depth @(racket z-jitter).
 The new points will lie within [@(racket x-min), @(racket x-max)] etc. if these bounds are non-@(racket #f).
+
+Note that adding random noise to data, via jittering or otherwise, is usually a bad idea.
+See the documentation for @(racket points) for examples where jittering may be appropriate.
 }
 
 @defproc[(vector-field3d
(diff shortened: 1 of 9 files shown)


plot-doc/plot/scribblings/renderer3d.scrbl
```diff
@@ -51,6 +54,10 @@ For example, a scatter plot of points sampled uniformly from the surface of a sp
                     
                     (plot3d (points3d (map vector xs ys zs) #:sym 'dot)
                             #:altitude 25)]
+
+When @(racket x-jitter), @(racket y-jitter), or @(racket z-jitter) is non-zero,
+ each point @(racket p) is translated to a random location inside a rectangle centered at @(racket p) with width @(racket x-jitter), height @(racket y-jitter), and depth @(racket z-jitter).
```
stamourv: s/rectangle/box/ ?

stamourv:
I think the docs should mention/discuss use cases. Otherwise, it's not clear when one might want this.
You may also want to hint at when it's even "ok" to do that. Randomly moving data points is (usually) not semantics-preserving! :)

bennn:
Good points, thanks

samth:
This blog post has a nice example of jitter use: http://kieranhealy.org/blog/archives/2015/02/03/another-look-at-the-california-vaccination-data/

bennn:
Should I add a margin note for this? (and maybe, [Typing with Pleasure](https://pavelfatin.com/typing-with-pleasure/))

samth:
Sounds like a good idea.

bennn:
Removed the extreme caution & added the external links.

stamourv:
LGTM

PR #12, opened 2016-01-28
Add #:samples to type for area-histogram

bennn:
The `#:samples` argument was implemented & documented, but not part of the type.

PR also adds a check for empty bins when making the `area-histogram`.
Empty bins are likely to happen `(- samples (length bin-bounds))` is small.

For example (using the `bin-samples` function from `plot-lib/plot/private/common/utils.rkt`):
```
(define bin-bounds '(0 9 10))
(define samples 4)
(define x* (linear-seq 0 10 samples
                       #:start? #f #:end? #f))

(displayln (bin-samples bin-bounds x*))
;; No samples fall in range [9, 10)
;; ((5/4 15/4 25/4 35/4) ())
```

Before this PR, `area-histogram` would raise a `division by zero` error


stamourv:
LGTM

PR #14, opened 2016-03-15
Rewrote and commented dead code

dkempe:
I wrote a comment associated with these changes in the original repository. This pull request implements the changes I referenced.

### DIFF
--- a/plot-lib/plot/private/common/math.rkt
+++ b/plot-lib/plot/private/common/math.rkt
@@ -146,11 +146,12 @@
 ;; ===================================================================================================
 ;; Vectors
 
+;; The final case was dead code and has been deleted. There are no instances of
+;; vector-andmap thattake more than two vectors as arguments.
 (:: vector-andmap
    (All (A B C ...)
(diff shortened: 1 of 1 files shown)


bennn:
This reminds me how the plot library needs continuous integration. I think I'll try adding that this weekend.

stamourv:
If this is dead code, we should just delete it, not comment it out.
If we do need it later, we can find it in the git history.

dkempe:
The problem with this approach is that the code is incorrect AND dead. If we just delete it and find that we need it later, we will revert to the code with out-of-scope vector references. I can submit two pull requests, one which fixes the code and one that deletes it, but they would need to be pulled in that order.

stamourv:
That makes sense.
Pull requests can have multiple commits in the, so only one is necessary.
IIRC, you can also force-push to this branch to update this PR, so you don't even need to create a new one.
Thanks for looking into this!

bennn:
Finally merged. Thanks again for working on this pull request.

87a0d2bf59cbc9458ae6ea874bdd24652fec4239
6ca968b43e4f7fa29a06942a13ca5516b45a4f33

PR #16, opened 2016-04-26
Fix _spec_ to match apply-jitter

bennn:
EDIT: docs now say that total deviation is 2x the amount given for `x-jitter` etc

### DIFF
--- a/plot-doc/plot/scribblings/params.scrbl
+++ b/plot-doc/plot/scribblings/params.scrbl
@@ -202,7 +202,7 @@ Used as default keyword arguments of @racket[points] and @racket[points3d].
 When any of @racket[x-jitter], @racket[y-jitter], or @racket[z-jitter] are non-zero,
  @racket[points] and @racket[points3d] will produce points randomly translated from their
  original position along the x, y, or z axis, respectively.
-For instance, if each parameter is set to 1, then @racket[points '(0 0)] will produce a random point
+For instance, if each parameter is set to 0.5, then @racket[points '(0 0)] will produce a random point
  in a square of area 1 centered at @racket['(0 0)].
 Likewise @racket[points3d] will make a random point within a unit cube centered at @racket['(0 0 0)].
 }
--- a/plot-doc/plot/scribblings/renderer2d.scrbl
+++ b/plot-doc/plot/scribblings/renderer2d.scrbl
@@ -80,22 +80,26 @@ The @racket[#:sym] argument may be any integer, a Unicode character or string, o
 @racket[known-point-symbols].
 Use an integer when you need different points but don't care exactly what they are.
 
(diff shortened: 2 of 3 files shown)


bennn:
Now the docs show:
screen shot 2016-04-26 at 06 03 31

stamourv:
Does it make more sense to specify the jitter in terms of the size of the rectangle (as the docs did, but the implementation didn't), or in terms of the maximum deviation from the original value (the other way around). The latter feels more intuitive to me, but given that I only have a vague sense of what that feature is for, take that with a pinch of salt. :)

bennn:
Either one is fine to me, but rectangle size seems easier to explain in the docs and less work to change in this PR.

samth:
What do other libraries with this feature do?

bennn:
R matches our implementation (not our docs) -- "total spread is twice the amount specified". I'll change this PR.
- http://docs.ggplot2.org/current/geom_jitter.html
- http://docs.ggplot2.org/current/position_jitter.html

bennn:
@stamourv just fyi here's a recent case where this feature was useful

We were suspicious that running the same benchmarks on multiple cores was giving very different results from single core measurements, so we ran a few versions of a program (numbered 0 - 15 on the x-axis) 30 times both sequentially and in parallel. Each green dot is 1 run of 1 version; all green dots were run sequentially on a dedicated core. Red dots are 1 run of 1 version but each version is run in parallel. The y-axis has the absolute runtimes.

shuffle-4-exact


Applying jitter helps see the actual distribution of the running times. I like this much better than using a boxplot or violin plot.

stamourv:
Given the new behavior, would it be better to have the docs written in terms of distance from the true value, as opposed to size of the rectangle?

rfindler:
@bennn can you explain how you've used jitter to produce this plot?

bennn:
@stamourv I don't like the `2*` factors, but I still think rectangles are the simplest, shortest way to explain what the parameters do. (Else idk, "e.g. `'(0 0)` is moved at most 1 unit along either axis" ?)

@rfindler the data had integer-valued x-coordinates: 60 points at x=0, 60 points at x=1, etc. I called `(points ... #:x-jitter 0.5)` to make it easier to see variations in the points' y-positions.

rfindler:
The smallest y coordinate on the plot appears to be 500? I'm confused.

bennn:
Yes, the y-coordinates seem to range between 400ms and 2000ms.

Here is the same plot without any jitter. Does this help the confusion?
ack

rfindler:
Yes, completely. Thanks!

stamourv:
@bennn: I don't have strong feelings either way re docs. Go with what you feel is best. LGTM otherwise.

bennn:
(Ran it by 2 other people, took the rectangle out of `points3d`'s explanation. Thanks for the feedback!)

Issue #18, opened 2016-06-07
Add type for categorical values


bennn:
[discrete-histogram](http://docs.racket-lang.org/plot/renderer2d.html?q=discrete-histogram#%28def._%28%28lib._plot%2Fmain..rkt%29._discrete-histogram%29%29) etc. take a variable called `cat-vals` as their first argument. Right now this can be a sequence of vectors or a sequence of lists, but as [Sean Kemplay](https://groups.google.com/forum/#!topic/racket-users/neFlqruebew) points out this type isn't really compatible with hashtables. And it would be nice to be able to call `discrete-histogram` on a hashtable.

I'm thinking the easiest fix is to give plot a type `Cat-Vals A` that would include the current sequences and also hashtables. Then `discrete-histogram3d`'s input would be a `Cat-Vals (Cat-Vals (U Real ivl #f))` and "categorical values" would have their own spot in the docs.

(I'm planning to set up a PR later this afternoon)

bennn:
I am losing the fight against the typechecker. Anyone know how to implement a function this signature?

```
(: f (-> (U (HashTable Symbol Natural)
            (Sequenceof (List Symbol Natural)))
         (Listof (Pairof Symbol Natural))))
```

I can't pick apart the types in the union. `hash?` and `sequence?` are too weak.

samth:
No, that's not possible, because hashes and sequences overlap.

bennn:
ok

samth:
Actually, I think `hash?` should be stronger here, since hashes are two-valued sequences and thus the types can't overlap. Maybe we can fix that.

PR #19, opened 2016-06-08
hash->cat-vals conversion function

bennn:
A possible fix to #18 that passes the typechecker.


bennn:
Use `(hash-map list H)` or `(in-values-sequence (in-hash H))` instead

PR #20, opened 2016-09-02
doc: use rectangles in  example

bennn:
Instead of squares, use rectangles in the example for the `rectangles` 2d renderer.
This way, it's easier to tell how each `ivl` affects the plot.

Also replaces "list" & "vector" with "sequence", to match the newer type signature.


stamourv:
LGTM

PR #22, opened 2017-05-11
doc: add the #:alpha argument for hrule and vrule

bennn:
Just found myself wanting a partly-transparent vertical line.
`#:alpha` wasn't in the docs. Bummer. Let's see about adding it.
Lo and behold 'twas already implemented.

I thought about adding a test, but wasn't sure if I could just append to `plot2d-tests.rkt`


bennn:
Ok the [release checklist](https://github.com/racket/racket/wiki/Release-Checklist) just says the plot tests are `raco test -c plot/tests` so I will append a small test.

PR #27, opened 2018-01-15
Provide plot3d interface unsafely to avoid excessive overhead.

samth:
These three provided functions don't take any higher-order arguments and their
results are actually implemented in untyped code, so there should not be a
risk of segfaults, plus they check their first-order arguments. Additionally,
the result contracts have not been checked in past releases due to a bug in
contract generation, increasing confidence that this is safe.

Fixes racket/gui#86. Related to racket/typed-racket#599.

I hope to merge this for 6.12 to fix the regression noted in
racket/gui#86.

### DIFF
--- a/plot-gui-lib/plot/private/gui/plot3d.rkt
+++ b/plot-gui-lib/plot/private/gui/plot3d.rkt
@@ -12,11 +12,12 @@
          plot/private/no-gui/plot3d
          plot/private/no-gui/plot3d-utils
          plot/private/no-gui/utils
-         "lazy-snip-typed.rkt")
+         "lazy-snip-typed.rkt"
+         typed/racket/unsafe)
 
-(provide plot3d-snip
-         plot3d-frame
-         plot3d)
+(unsafe-provide plot3d-snip
+                plot3d-frame
+                plot3d)
 
 ;; ===================================================================================================
 ;; Plot to a snip
--- a/plot-lib/plot/private/no-gui/plot3d-utils.rkt
+++ b/plot-lib/plot/private/no-gui/plot3d-utils.rkt
@@ -13,14 +13,16 @@
 
 (provide (all-defined-out))
 
-(: get-renderer-list (-> (Treeof (U renderer3d nonrenderer)) (Listof renderer3d)))
+(: get-renderer-list (-> Any (Listof renderer3d)))
 (define (get-renderer-list renderer-tree)
   (cond [(list? renderer-tree)  (append* (map get-renderer-list renderer-tree))]
         [(nonrenderer? renderer-tree)
          (match-define (nonrenderer bounds-rect bounds-fun ticks-fun) renderer-tree)
          (list (renderer3d bounds-rect bounds-fun ticks-fun #f))]
+        [(renderer3d? renderer-tree)
+         (list renderer-tree)]
         [else
-         (list renderer-tree)]))
+         (raise-argument-error 'get-renderer-list "(or/c list? nonrenderer? renderer3d?)" renderer-tree)]))
 
 (: get-bounds-rect (-> (Listof renderer3d)
                        (U #f Real) (U #f Real)


samth:
cc @racket/release

rfindler:
This seems risky, unless there has been considerable testing. Removing the types seems like only a little more effort (and worth it).

samth:
Removing the types from what? Just removing the types from that one file would just move the contract boundary somewhere else. Removing the types from all of plot would be a far more drastic change than this one.

stamourv:
Re being tested: as @samth said, this code already has a history of running without checking, without issue. That, and the checks in the code proper make me fairly confident.

I also agree 100% with @samth's latest comment. Given the track record of contract generation for this library, moving the contract boundary somewhere else seems risky. Riskier than manually disabling some, at least.

Merging this into 6.12 is the right choice, IMO.

rfindler:
I would feel much better a second person explicitly did an audit of the assertions in Sam's original comments.

stamourv:
@rfindler: I agree. @bennn: You have some familiarity with that part of the codebase, is that something you could do?

bennn:
Yes, I'm planning to give this a review in the next few hours.

Here are the "assertions" I see in Sam's original comments. Let me know if you think I'm missing any:
- the 3 functions don't take higher-order arguments
- the 3 functions check their first-order arguments
- the results of the 3 functions are implemented in untyped code
- this PR is safe against segfaults

samth:
Just as a guide, everything in this file is really going through `make-3d-plot-snip`, which is imported via `lazy-snip-typed.rkt` where it is given a type without a contract using TR internals. It's defined in `lazy-snip-untyped.rkt` where it just calls `make-3d-plot-snip` from `snip3d.rkt`. That file is implemented using `#lang racket/base`, and depends on `snip.rkt` which is also in `#lang racket/base`.

rfindler:
Those are the assertions I see. Thanks, @bennn !

rfindler:
(I just looked up the word "assertion" and I see it doesn't mean exactly what I used it to mean. What I meant was "claim that can be proved or refuted".)

samth:
Thanks, @bennn.

For thinking about the release, I'll just lay out the options and some pro/cons that I see.

1. Do nothing. Pro: no change risk. Con: 3d plot snips are very slow, as reported in racket/gui#86.
2. Take this patch. Pro: fixes problem; simple easy-to-audit change; no impact outside this one file. Con: it's possible that misuse of one of the functions from `plot3d.rkt` could trigger incorrect behavior up to and including memory corruption, although I believe that this is not actually possible.
3. Do what we did for 6.11, which is revert a change to TR made by @bennn. Pro: fixes the problem; we know that it worked last time. Con: adds unsafety to many TR-using programs. 

All the other changes I've thought of seem much larger and more risky for the 6.12 release.

jeapostrophe:
I've looked at the code too and support the change.

bennn:
Short version: I found one issue with this PR. It should be easy for someone who understands the TR optimizer to fix it.

samth:
What if you just change the annotation for the domain of that function to
Any?

bennn:
Changing the domain fixes the segfault. Here's the code I used:

```
(: get-renderer-list (-> Any (Listof renderer3d)))
(define (get-renderer-list renderer-tree)
  (cond [(list? renderer-tree)  (append* (map get-renderer-list renderer-tree))]
        [(nonrenderer? renderer-tree)
         (match-define (nonrenderer bounds-rect bounds-fun ticks-fun) renderer-tree)
         (list (renderer3d bounds-rect bounds-fun ticks-fun #f))]
        [(renderer3d? renderer-tree)
         (list renderer-tree)]
        [else
         (raise-argument-error 'get-renderer-list "(or/c list? nonrenderer? renderer3d?)" renderer-tree)]))
```

And the new error message for `(plot3d #f)` is:

```
get-renderer-list: contract violation
  expected: (or/c list? nonrenderer? renderer3d?)
  given: #f
```

samth:
Great. I think you can update the pull request, but if not I'll do it very
soon.

stamourv:
Thanks @bennn.

I agree that changing the type to `Any` would help.

What about (also) adding `#:no-optimize` to all the typed files in the `plot` collection? I doubt that would affect performance much, and it would limit the damage that could result from disabling the types to Racket errors in the internals, rather than segfaults. That should be a fairly conservative chance, I believe.

stamourv:
Minimal change, with @bennn's amendment, is good enough for me for this release.

samth:
I've merged this, and I see @stamourv has created the release branch with this commit. Thanks everyone (especially @bennn) for helping out by reviewing my changes.

PR #28, opened 2018-01-17
plot3d regression tests

bennn:
This PR adds regression tests for f53d103d37827ea83b41015da59fcd519a9345e0 , to check that the plot3d functions don't segfault on bad arguments.

While testing I realized I made a mistake yesterday (#27) --- plot's parameters are **not** always guarded with a contract. The parameters **do** seem safe against segfaults. But a bad value can give a bad error message, so this PR explicitly checks all keyword arguments to plot3d functions.

Also adds a `plot-file-format/c` contract. (Not necessary, but I think it's a good idea.)

(cc @stamourv I am not suggesting we merge this pull request into the release)

### DIFF
--- a/plot-gui-lib/plot/private/gui/plot3d.rkt
+++ b/plot-gui-lib/plot/private/gui/plot3d.rkt
@@ -20,6 +20,11 @@
                 plot3d-frame
                 plot3d)
 
+(require/typed plot/utils
+  (anchor/c (-> Any Boolean))
+  (plot-color/c (-> Any Boolean))
+  (plot-file-format/c (-> Any Boolean)))
+
 ;; ===================================================================================================
 ;; Plot to a snip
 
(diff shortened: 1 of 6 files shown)


bennn:
ping, any thoughts about my changes to the documentation?
https://github.com/racket/plot/pull/28/commits/bd5986dcb9e39c2547d11e4c7555dddf6bb93470

Otherwise, I'll plan to merge this next weekend.

plot-gui-lib/plot/private/gui/plot3d.rkt
```diff
@@ -189,8 +196,8 @@
                 #:z-min [z-min #f] #:z-max [z-max #f]
                 #:width [width (plot-width)]
                 #:height [height (plot-height)]
-                #:angle [angle (plot3d-angle)]
-                #:altitude [altitude (plot3d-altitude)]
+                #:angle [angle #f]
+                #:altitude [altitude #f]
```
bennn: Note: if these are not `#false`, then the `#:az` and `#:alt` arguments are unused --- even though L246 and others try to use them.

stamourv:
Changes to the docs looks good. Haven't had a thorough look at the rest, though.

PR #29, opened 2018-01-19
Provide plot2d interface unsafely to avoid excessive overhead

bennn:
https://github.com/racket/drracket/issues/185

cc @racket/release


stamourv:
Do the arguments that made the change safe in the 3d case apply here too?

bennn:
I think so, but I haven't looked carefully yet. Should I do that ASAP, or can it wait until later tonight?

stamourv:
Later tonight is fine.

jbclements:
Many thanks!

bennn:
Yep, `plot-snip` `plot-frame` and `plot` appear safe to provide unsafely.

PR #30, opened 2018-01-19
Convert to no-check

stamourv:
This library has consistent performance problems due to TR's contract generation.

This PR removes type checking, to avoid contract generation altogether.

This is a fairly crude job, but one which appears to be functional. More thorough testing is welcome.

One limitation is that in the absence of types and with no contracts, error messages will be pretty bad. Adding contracts directly should solve that, but I haven't done it.


mflatt:
I support this direction in the short run (though not as short as the v6.12 release; we want more minimal changes there).

In the long run, of course, we're all pulling for Typed Racket plus good performance plus good build times plus high reliability. But the build time/space part has been especially elusive, in addition to some recent reliability problems, so I think it makes sense to back off TR for a little while.

Besides error messages, are there any safety issues with just turning off types and the contracts that they imply?

rfindler:
I agree completely with Matthew in that I'm pulling for TR in the long run, but we need some short-run patches (or else, IMO, we won't have a chance to make the long run). I say this being aware that this is a backwards incompatible change (although I think we should try to look into pkg-server-listed pkgs that depend on this and offer a warning or a pull request when we find them).

As for safety issues, I think the worst that could happen is ill-timed errors (in GUI threads like in the DrRacket interactions window) or garbled plots. Clearly things we want to try to avoid!

samth:
I'm against this change, for a number of reasons.

As background, I think it's important to be clear that the problems that plot has had are not about typechecking or Typed Racket in general, but about class and object contracts. Unfortunately, contracts for `racket/class` have consistently turned out to be more expensive that people have hoped, in runtime performance, in expansion size, and in compile time. Although I haven't worked on their implementation, the sense I get is that people have tried to bring down the overhead but there are hard design and implementation problems that we haven't solved yet.

In other cases in Racket where the cost of contracts have been a significant problem, we have not taken contracts out of the code, but worked out other approaches (such as `syntax/private/id-table`). I think this recognizes the significant software engineering benefits that we believe contracts provide. Here, the proposal is to remove not just contracts, but also types, which I think we agree provide some additional benefit. So I think this would be a significant loss, and an approach not in keeping with how we've done these things in the past when running into similar issues.

Some other comments:

1. `plot` makes use of the `math` library, which is also typed, and which can have performance issues when used from untyped code. Do we know that this PR won't cause other performance regressions in `plot`? As we've seen, it can be very hard to detect these without user testing.

2. Using `typed/racket/no-check` means that the type annotations are effectively comments, but as we know, comments in source code are bound to get out of sync with the actual code. I worry that leaving the types in without checking them will result in significant confusion down the road.

3. Some of the compile-time/zo-size problems are alleviated (but not eliminated) by racket/typed-racket#630, which I really hope to get merged soon. Paternity leave has drastically limited my time to work on these things at the moment, but it's basically ready and makes a big difference. This of course does not improve any future runtime performance problems.

4. `plot` uses a variety of tricks to get around the type system in places. Does using `no-check` cause any of them to break?

Finally (and this is last intentionally) Typed Racket (and Racket more broadly) are part of many people's research, and the `plot` library has been helpful in pushing that forward. That's both a concern for researchers as well as for the future development of Racket and Typed Racket, which are better because they've had this library pushing them forward.

samth:
We did other things to remove these contracts.

PR #32, opened 2018-01-30
add interactive overlays to plot-snip% instances

alex-hhh:
The plot-snip% class has been extended with the following methods:

* set-hover-callback -- sets a callback that receives the snip and mouse
  location in plot coordinates

* add-vrule-overlay, add-hrule-overlay, add-xrule-overlay -- adds a vertical
  line, horizontal line or both to the plot

* add-mark-overlay -- adds a marker (a circle) and a label

* add-general-overlay -- provides a function to draw arbitrary things on the
  plot

* clear-overlays, refresh-overlays -- helper functions

This allows overlays that update as the mouse moves over the plot.

### DIFF
--- a/plot-gui-lib/plot/private/gui/lazy-snip-types.rkt
+++ b/plot-gui-lib/plot/private/gui/lazy-snip-types.rkt
@@ -9,9 +9,10 @@
   (-> (Instance Bitmap%)
       Plot-Parameters
       (-> Boolean Rect Positive-Integer Positive-Integer
-          (Values (Instance Bitmap%) Rect (-> Rect Rect)))
+          (Values (Instance Bitmap%) (U #f (Instance 2D-Plot-Area%)) Rect (-> Rect Rect)))
       Rect
       Rect
+      (U #f (Instance 2D-Plot-Area%))
       (-> Rect Rect)
       Positive-Integer
       Positive-Integer
--- a/plot-gui-lib/plot/private/gui/lazy-snip-untyped.rkt
+++ b/plot-gui-lib/plot/private/gui/lazy-snip-untyped.rkt
@@ -13,10 +13,10 @@
 
 (define (-make-2d-plot-snip
          init-bm saved-plot-parameters
-         make-bm plot-bounds-rect area-bounds-rect area-bounds->plot-bounds width height)
+         make-bm plot-bounds-rect area-bounds-rect area area-bounds->plot-bounds width height)
   (make-2d-plot-snip
    init-bm saved-plot-parameters
-   make-bm plot-bounds-rect area-bounds-rect area-bounds->plot-bounds width height))
+   make-bm plot-bounds-rect area-bounds-rect area area-bounds->plot-bounds width height))
 
 (define (-make-3d-plot-snip
          init-bm saved-plot-parameters
--- a/plot-gui-lib/plot/private/gui/plot2d.rkt
+++ b/plot-gui-lib/plot/private/gui/plot2d.rkt
@@ -57,7 +57,7 @@
        (define bounds-rect (get-bounds-rect renderer-list x-min x-max y-min y-max))
        
        (: make-bm (-> Boolean Rect Positive-Integer Positive-Integer
-                      (Values (Instance Bitmap%) Rect (-> Rect Rect))))
+                      (Values (Instance Bitmap%) (U #f (Instance 2D-Plot-Area%)) Rect (-> Rect Rect))))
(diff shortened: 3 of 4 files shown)


alex-hhh:
Hi There,

I created this PR as suggested by Matthias Felleisen on the Racket Users Group.  I don't think this is ready to be merged yet, but I would like to receive some feedback on the implementation approach.

I prepared a GIST with an example usage script, along with some example images from my application (which actually prompted me to implement this feature):

https://gist.github.com/alex-hhh/64f8af9c4bd5b5c65ada62dd9157e938

The obvious thing that is missing is documentation, and perhaps some contracts/types on the new methods, however, I would also like to be sure that the proposed API changes are actually adequate for my application.  In the next few days (weeks?) I will update my application to add overlays to histogram and scatter plots and see how that works out.

Thanks,
Alex.

alex-hhh:
Regarding your first point, I am not sure how feasible would it be to make a generic "extended" interface for `2d-plot-snip%` -- the ZOOM-UNZOOM functionality cannot be implemented externally as it used too much of the `2d-plot-snip%` internals.

Also, the `2d-plot-snip%` needs the overlay API in addition to a `set-mouse-event-handler` method, as sometimes overlays are added not in response to mouse events.  In my blog post I showed two examples:

* the linked plots where a vertical line is added to several plots when the mouse hovers over one of them only
  
* highligting segments on a plot when they are selected in some other part of the GUI.

----

Regarding using `renderer2d` objects for the overlays: I am not sure if it could be done or not, but here are the things I considered before making the current design decisions

## The renderer tree cannot be accessed directly or modified

The current `2d-plot-snip%` object does not have access to the renderer tree. Instead, it only has access to the function `make-bm` which can create a new bitmap image with a new plot area.  This function is passed in as a closure in `plot-snip`("snip2d.rkt" file).  As such we cannot modify the renderer tree from inside a `2d-plot-snip%` object, at least not without significant changes to the code.

## It is undesirable to modify the main renderer tree

There would be a severe performance drop if the overlays would be added to the main render tree and the whole thing re-rendered.  Overlays have to be drawn after every mouse move event and this is a lot of re-draws.

In my use case, it takes about 2 - 3 seconds to render the plots for complex, but real world data.  The speed of the overlay implementation relies on the that the actual plot is stored as a bitmap image and can be quickly re-drawn, followed by drawing the overlays.

My previous "highlight segment" implementation relied on modifying the renderer tree to add `function-interval` renderer objects for the selected ranges and, while it worked, it was barely usable.

## Using a secondary renderer tree

We could use a secondary renderer tree for the overlays.  I am not sure how difficult would it be to make the secondary renderer tree line up on top of the main one.

This approach would also make label placement more complex: currently, labels are positioned such that they are always inside the plot area, and this happens automatically.  The `point` renderer, which has a similar function, does allow for a label, but makes the user specify its placement.  It is not possible to determine the position of the label from user code, such that the label is always inside the plot, as the user code has no access to the pixel dimensions of the plot area.

There is also the issue of drawing arbitrary things, as per `add-general-overlay`, which is not supported by the built in renderers.  I am using that feature to draw the badges on various plots, which are pict instances.

bennn:
Thanks for the answers.

I'm not sure why zoom/unzoom couldn't be implemented externally, but I'll try to figure out why for myself (this weekend).

Having some kind of overlay API makes sense. It also makes sense to not modify the main renderer tree and to not force the user to pick a label anchor (at least, `'auto` should be a valid anchor for an overlay version of plot's `point-label`).

I think it would be good to use a second renderer tree --- or, a second tree that's similar to the renderer tree built from the structs in this PR. Basically, I think plot's overlay API should look similar to plot's plot API, so users don't have to learn a brand new interface.

- - -

To summarize my thoughts on the implementation approach: (1) I dislike that there's so many changes in `snip2d.rkt` (new datatypes, lots of new methods), and (2) I strongly dislike the fact that the `add-X-overlay` methods use similar names as plot renderers in a dis-similar style. [[ Most of the reason for "strong dislike" is that we won't be able to change / remove these methods after they're merged without possibly breaking users' code ]]

alex-hhh:
I have added a new function, `set-overlay-renderers`, which allows using a renderer tree as overlays.  This was not very complex -- I was actually surprised by this, however it creates an artifact when the snip is only partially visible.  The gist below shows example usage as well as a screenshot of the problematic case.

https://gist.github.com/alex-hhh/2394d60e57953f82e363fc68fae877b7

alex-hhh:
Closing this pull request as it is superseded by #37

PR #33, opened 2018-02-08
Add 'auto Anchor and 'none Point-Sym for `point-label` and related functions

alex-hhh:
An `'auto` `Anchor` type can be used to position the label inside the plot, without needing to know the plot edges.  A `'none` `Point-Sym` will cause `point-label` to not show the current location on the plot, it will just show the label.

Example code:

```racket
#lang racket
(require plot)
    
(plot
 ;; Add labels on the edges of the plot area.  Each label will be positioned
 ;; so it is inside the plot area
 (list (point-label (vector 0 10) #:anchor 'auto)
       (point-label (vector 0 -10) #:anchor 'auto)
       ;; Side labels have no corresponding point
       (point-label (vector 10 0) #:anchor 'auto #:point-sym 'none)
       (point-label (vector -10 0) #:anchor 'auto #:point-sym 'none))
 #:x-min -10 #:x-max 10
 #:y-min -10 #:y-max 10)
```

Results in:
auto-anchor

This pull request is a step towards interactive overlays based on renderer trees as suggested by @bennn in #32 .  The feature will be most useful to add point labels interactively, but it can be useful independently of that.

### DIFF
--- a/plot-doc/plot/scribblings/contracts.scrbl
+++ b/plot-doc/plot/scribblings/contracts.scrbl
@@ -102,7 +102,7 @@ The contract for the @(racket #:sym) arguments in @(racket points) and @(racket 
         'circle7           'circle8          'bullet
         'fullcircle1       'fullcircle2      'fullcircle3
         'fullcircle4       'fullcircle5      'fullcircle6
-        'fullcircle7       'fullcircle8)]{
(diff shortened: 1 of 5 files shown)


alex-hhh:
Closing this as the changes from it were merged as part of #34.

PR #34, opened 2018-02-10
 Add `point-pict` 2d renderer, similar to `point-label`

alex-hhh:
The `point-pict` function adds a point to the graph and specifies a pict object to be displayed as the label.  This is similar to the `point-label` function, which can only show a string as the label.

`function-pict`, `inverse-pict`, `parametric-pict` and `polar-pict` are added similar to their *-label counterparts.

Sample code:

```racket
#lang racket
(require plot pict)

(define blue-fish (standard-fish 40 15 #:color "lightblue"))
(define red-fish (standard-fish 40 15 #:color "salmon"))

(define (f t) (vector (cos t) (sin t)))

(plot
 (list (point-pict (vector 0 10) blue-fish #:anchor 'auto)
       (point-pict (vector 0 -10) blue-fish #:anchor 'auto #:point-size 15)
       (point-pict (vector 10 0) blue-fish #:anchor 'auto #:point-sym 'none)
       (point-pict (vector -10 0) blue-fish #:anchor 'auto #:point-sym 'none)
       (parametric f 0 (* 2 pi))
       (parametric-pict f (/ pi 2) red-fish #:anchor 'bottom )
       (function sin)
       (function-pict sin -5 red-fish #:anchor 'left)
       (polar-pict (lambda (a) (* 8 (sin a))) (/ pi 4) red-fish)
       (inverse-pict asin 1 red-fish))
 #:x-min -10 #:x-max 10
 #:y-min -10 #:y-max 10)
```

Produces:

point-snip-sample

This pull request is built on top of #33, and it contains its commits, only the last commit implements this functionality.

PR #35, opened 2018-02-11
test: add tests for 'auto, point-pict, etc

bennn:
Add tests from #33 and #34 to the repo.

@alex-hhh I made the second test (from #34) a little simpler because the test from #33 already plots 2 `'none` points. Let me know if you think we should make other changes, or add other tests.


alex-hhh:
Hi @bennn , I'm not sure how these tests work?  Are they automated somehow or it relies simply on running the code and the user looking at the plots?

Otherwise, the changes do cover most of the use cases for the new features.

alex-hhh:
You can also add the tests from #36 to this pull request.

PR #36, opened 2018-02-12
Extend `rectangles` renderer to accept open intervals

alex-hhh:
When #f is specified to one of the interval bounds of the `rectangles`
renderer, the rectangle will extend to the edge of the plot area in the
direction of that bound.  Example:

```racket
#lang racket
(require plot)
(plot (rectangles (list (vector (ivl 2 3) (ivl #f #f))
                        (vector (ivl #f #f) (ivl 2 3))
                        ;; Vertical extent from 0 to the top
                        (vector (ivl 5 6) (ivl 0 #f)))
                  #:line-style 'transparent
                  #:alpha 0.5)
      #:x-min -10 #:x-max 10
      #:y-min -10 #:y-max 10)
```

Implementation note: the rectangles renderer code used to compute the bounds
of the plot based on the bounds of the rectangles, this code is unnecessary
and complicates the case where the intervals are open-ended.  It was removed.

### DIFF
--- a/plot-doc/plot/scribblings/renderer2d.scrbl
+++ b/plot-doc/plot/scribblings/renderer2d.scrbl
@@ -550,9 +550,9 @@ For example, the canonical saddle, with its gradient field superimposed:
 Returns a renderer that draws rectangles.
 
 The rectangles are given as a sequence of sequences of intervals---each inner
-sequence defines the bounds of a rectangle.  Any of the bounds can be #f, in
-which case the rectangle extents to the edge of the plot area in the
-respective direction.
+sequence defines the bounds of a rectangle.  Any of the bounds can be
+@racket[-inf.0] or @racket[+inf.0], in which case the rectangle extents to the
+edge of the plot area in the respective direction.
 
 For example,
 @interaction[#:eval plot-eval (plot (rectangles (list (vector (ivl -1 0) (ivl -1 1))
--- a/plot-doc/plot/scribblings/renderer2d.scrbl
+++ b/plot-doc/plot/scribblings/renderer2d.scrbl
@@ -548,7 +548,13 @@ For example, the canonical saddle, with its gradient field superimposed:
           [#:label label (or/c string? #f) #f]
           ) renderer2d?]{
 Returns a renderer that draws rectangles.
-The rectangles are given as a sequence of sequences of intervals---each inner sequence defines the bounds of a rectangle. For example,
+
+The rectangles are given as a sequence of sequences of intervals---each inner
+sequence defines the bounds of a rectangle.  Any of the bounds can be #f, in
+which case the rectangle extents to the edge of the plot area in the
+respective direction.
+
+For example,
 @interaction[#:eval plot-eval (plot (rectangles (list (vector (ivl -1 0) (ivl -1 1))
                                                       (vector (ivl 0 2) (ivl 1 2)))))]
 }
--- a/plot-doc/plot/scribblings/utils.scrbl
+++ b/plot-doc/plot/scribblings/utils.scrbl
@@ -370,7 +370,7 @@ Returns @racket[#t] if every element of @racket[v] is @racket[rational?].
 @;----------------------------------------------------------------------------------------------------
 @subsection[#:tag "math.intervals"]{Intervals and Interval Functions}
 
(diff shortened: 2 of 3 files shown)


alex-hhh:
open-rectangles

plot-lib/plot/private/plot2d/rectangle.rkt
```diff
@@ -59,24 +62,9 @@
...
-       (cond
-         [(or (empty? rxs) (empty? rys))  (renderer2d #f #f #f #f)]
-         [else
-          (let ([x-min  (if x-min x-min (apply min* rxs))]
-                [x-max  (if x-max x-max (apply max* rxs))]
-                [y-min  (if y-min y-min (apply min* rys))]
-                [y-max  (if y-max y-max (apply max* rys))])
-            (renderer2d (vector (ivl x-min x-max) (ivl y-min y-max)) #f default-ticks-fun
-                        (rectangles-render-proc rects color style line-color line-width line-style
-                                                alpha label)))]))]))
```
bennn: Removing this code breaks the example in the docs:

```
#lang racket/base
(require plot)
(plot (rectangles (list (vector (ivl -1 0) (ivl -1 1))
                        (vector (ivl 0 2) (ivl 1 2)))))
```

because plot is no longer able to infer the bounds.

- - -

I think the old code would be able to handle open-ended intervals if the condition on L71 was something like:

```
[(or (and (empty? x1s) (empty? x2s)) (and (empty? y1s) (empty? y2s)))  (renderer2d #f #f #f #f)]
```

Or maybe, this line can just be removed.

plot-doc/plot/scribblings/renderer2d.scrbl
```diff
@@ -548,7 +548,13 @@ For example, the canonical saddle, with its gradient field superimposed:
           [#:label label (or/c string? #f) #f]
           ) renderer2d?]{
 Returns a renderer that draws rectangles.
-The rectangles are given as a sequence of sequences of intervals---each inner sequence defines the bounds of a rectangle. For example,
+
+The rectangles are given as a sequence of sequences of intervals---each inner
+sequence defines the bounds of a rectangle.  Any of the bounds can be #f, in
```
bennn: `#f` should be `@racket[#f]`

alex-hhh:
I updated the change set to use `+/-inf.0` for the "extend to the edge of the plot" functionality -- this seems to show intent more clearly and it looks like `#f` was used for something else.  Looking at the `ivl` definition more closely, I'm not sure why ivl definitions allow `#f`, as rectangles that contain `#f` are simply not drawn...

```racket
#lang racket
(require plot)
(plot (rectangles (list (vector (ivl 2 3) (ivl -inf.0 +inf.0))
                        (vector (ivl -inf.0 +inf.0) (ivl 2 3))
                        ;; Vertical extent from 0 to the top
                        (vector (ivl 5 6) (ivl 0 +inf.0)))
                  #:line-style 'transparent
                  #:alpha 0.5)
      #:x-min -10 #:x-max 10
      #:y-min -10 #:y-max 10)
```

plot-doc/plot/scribblings/utils.scrbl
```diff
@@ -370,7 +370,7 @@ Returns @racket[#t] if every element of @racket[v] is @racket[rational?].
 @;----------------------------------------------------------------------------------------------------
 @subsection[#:tag "math.intervals"]{Intervals and Interval Functions}
 
-@defstruct[ivl ([min real?] [max real?])]{
+@defstruct[ivl ([min (or/c real? #f)] [max (or/c real? #f)])]{
```
bennn: Since `rectangles` is using `inf` values now, I propose un-doing this change to the docs.

[[ I have no idea why `ivl`s can contain `#false` internally. So as long as this PR is not adding a new API that allows `#false`, I think we better keep it a secret that `(ivl #false #false)` is allowed. ]]

bennn:
Looks good!

alex-hhh:
I reverted the changes to `utuls.scrbl`, although I'm not sure why do we need to hide from the user that the `ivl` allows `#f` members...

bennn:
I'm not sure if we need to hide it either, but since this pull request no longer needs to let users know that an `ivl` can hold `#f` values, I'm glad this pull request isn't the one to reveal that `#f` is allowed.

Issue #2, opened 2015-03-04
Segfault in plot tests


takikawa:
I tried to run the `plot/tests/typed/parameter-regression-tests.rkt` in DrRacket and get a segfault. It appears that just opening this file (also `plot-pict.rkt` in the same directory) and waiting for background compilation to finish will trigger a segfault.

@stamourv was also able to reproduce this on his machine.

ntoronto:
I can't reproduce this on 6.1.1.8--2015-01-31. I'll try updating and see what I get.

The fact that background compilation can set it off suggests a Typed Racket error, but I'll try to reproduce it before punting.

mflatt:
I can replicate it. I think the problem is in the macro expander.

ntoronto:
I can't replicate the crash, but I'm not getting any plot snips at all. For example, this program gives me two plots, a file, and a useless single-pixel I-don't-know-what-it-is:
```racket
#lang racket
(require plot)
(plot-pict (function sin -3 3))
(plot-bitmap (function sin -3 3))
(plot-file (function sin -3 3) "blah.png")
(plot-snip (function sin -3 3))
```

I'd suggest looking in "plot/plot-gui-lib/plot/private/gui/untyped.rkt", which defines untyped wrappers for lazily-required untyped code that creates snips, and "plot/plot-gui-lib/plot/private/gui/plot2.rkt", which imports those wrappers using `require/typed`.

mflatt:
I've pushed a repair for the crash

ntoronto:
I'll close this, then. The single-pixel-snip problem appears to be unrelated, and I've submitted a bug report for it.

PR #37, opened 2018-02-12
Update 2d-plot-snip% to allow interactive overlays.

alex-hhh:
Two methods are added to the 2d-plot-snip% (which is returned by the
`plot-snip` function):

* `set-mouse-event-callback` installs a user callback for handling mouse
  events from the plot snip.  The callback should be in the format:

  (lambda (snip mouse-event x y) ...)

  Where SNIP is the plot snip generating the event, MOUSE-EVENT is the event
  generated and X, Y are the location of the mouse in plot coordinates

* `set-overlay-renderers` adds a list of 2d renderers to be drawn on top of
  the plot.  This can be any valid 2d renderer.  Only one set of renderers is
  valid at a time, calling this function a second time, will replace the
  previous overlay renderers.  Calling the function with an #f argument will
  cause the overlay renderers to be removed.

Example usage:

```racket
#lang racket
(require plot pict)

(define fish (standard-fish 40 15 #:color "salmon"))

(define (on-hover snip event x y)
  (if (and x y)
      (let ((sx (sin x)))
        (send snip set-overlay-renderers
              (list (vrule x)
                    (hrule sx)
                    (point-label (vector x sx) "hello" #:anchor 'auto)
                    (point-pict (vector x y) fish #:anchor 'auto))))
      (send snip set-overlay-renderers #f)))

(define snip (plot-snip (function sin) #:x-min -10 #:x-max 10))
(send snip set-mouse-event-callback on-hover)
snip
```

### DIFF
--- a/plot-doc/plot/scribblings/renderer2d.scrbl
+++ b/plot-doc/plot/scribblings/renderer2d.scrbl
@@ -890,3 +890,99 @@ Returns a renderer that draws a labeled point on a polar function's graph.
           ) renderer2d?]{
 Returns a renderer that draws a point with a pict as the label on a polar function's graph.
 }
+
+@section[#:tag "2d-plot-snip-interactive-overlays"]{Interactive Overlays for 2D plots}
+
+An plot object returned by @racket[plot-snip] can be set up to provide
+interactive overlays, such as displaying the current value of the plot
+function at he mouse cursor.  Two methods are available on the returned
+@racket[snip%] object for this purpose.
+
+@defclass[2d-plot-snip% snip% ()]{
+
+An instance of this class is returned by @racket[plot-snip].
+
+@defmethod[(set-mouse-event-callback [callback (or/c plot-mouse-event-callback/c #f)]) any/c]{
+
+Set a callback function to be invoked with mouse events from the snip.  The
+callback is invoked with the actual snip object, the @racket[mouse-event%] and
+the X, Y position of the mouse in plot coordinates -- these are in the same
+coordinate system used by the renderers in the plot.  The X, Y values are
+@racket[#f] when the mouse is outside the plot area (for example, when the
+mouse is over the axis area).
+
+When a callback is installed, the default zoom functionality of the plot snips
+is disabled.  This can be restored by calling
+@racket[set-mouse-event-callback] with a @racket[#f] argument.
+
+}
+
+@defmethod[(set-overlay-renderers [renderers (or/c #f (treeof renderer2d?))]) any/c]{
+
+Set a list of renderers (or more generally, a tree of renderers) to be drawn
+on top of the existing plot.  This can be any combination of 2D renderers, but
+it will not be able to modify the axes or the dimensions of the plot area.
+Only one set of overlay renderers can be installed, calling this method a
+second time will replace the previous overlays.  Specifying @racket[#f] as the
+renderers will cause overlays to be disabled.
+
+}
+}
+
+@defthing[plot-mouse-event-callback/c contract? #:value (-> (is-a?/c snip%)
+                                                       (is-a?/c mouse-event%)
+                                                       (or/c real? #f)
+                                                       (or/c real? #f) any/c)]{
+A contract for callback functions passed to @racket[set-mouse-event-callback].
+}
+
+
+For example, if you evaluate the code below in DrRacket, the resulting plot
+will show a vertical line tracking the mouse and the current plot position is
+shown on a label.  This is achieved by adding a mouse callback to the plot
+snip returned by @racket[plot-snip].  When invoked, the mouse callback will
+add a @racket[vrule] at the current X position and a @racket[point-label] at
+the current value of the plotted function.
+
+@examples[#:eval plot-eval
+(require plot)
+(define snip (plot-snip (function sin) #:x-min -5 #:x-max 5))
+(define (mouse-callback snip event x y)
+   (if (and x y)
+       (send snip set-overlay-renderers (list (vrule x) (point-label (x (sin x)))))
+       (send snip set-overlay-renderers #f)))
+(send snip set-mouse-event-callback mouse-callback)
+snip]
+
+Here are a few hints for adding common interactive elements to racket plots:
+
+@itemlist[
+
+@item{@(racket hrule) and @(racket vrule) renderers can be used to draw
+horizontal and vertical lines that track the mouse position}
+
+@item{The @(racket rectangles) renderer can be used to highlight a region on
+the plot.  You can also specify @racket[-inf.0] and @racket[+inf.0] as limits
+for one of the dimensions, to extend the rectangle all the way to the edge of
+the plot in that direction.  You will need to specify a non zero value for the
+@racket[#:alpha] parameter, so the rectangle is transparent and the plot
+underneath it is visible. For example, to highlight a vertical region between
+@racket[xmin] and @racket[xmax], you can use:
+
+@racket[
+(rectangles (list (vector (ivl xmin xmax) (ivl -inf.0 +inf.0)))
+            #:alpha 0.2)]
+}
+
+@item{A @(racket point-label) renderer can be used to add a point with a
+string label to the plot.  To add only the label, use @racket['none] as the
+value for the @racket[#:point-sym] argument.}
+
+@item{A @(racket point-pict) renderer can be used to add a point with an
+attached @racketmodname[pict] instead of a string label.  This can be used to
+draw fancy labels (for example with rounded corners), or any other type of
+graphics element.}
+
+@item{A @(racket points) renderer can be used to mark specific locations on
+the plot, without specifying a label for them}
+]
--- a/plot-doc/plot/scribblings/renderer2d.scrbl
+++ b/plot-doc/plot/scribblings/renderer2d.scrbl
@@ -893,56 +893,19 @@ Returns a renderer that draws a point with a pict as the label on a polar functi
 
 @section[#:tag "2d-plot-snip-interactive-overlays"]{Interactive Overlays for 2D plots}
 
-An plot object returned by @racket[plot-snip] can be set up to provide
-interactive overlays, such as displaying the current value of the plot
-function at the mouse cursor.  Two methods are available on the returned
-@racket[snip%] object for this purpose.
+@declare-exporting[plot/snip]
+@defmodule*/no-declare[(plot/snip)]
 
-@defclass[2d-plot-snip+c% snip% ()]{
+A plot @racket[snip%] object returned by @racket[plot-snip] can be set up to
+provide interactive overlays.  This feature can be used, for example, to show
+the current value of the plot function at the mouse cursor.
 
-An instance of this class is returned by @racket[plot-snip].
-
-@defmethod[(set-mouse-event-callback [callback (or/c plot-mouse-event-callback/c #f)]) any/c]{
-
-Set a callback function to be invoked with mouse events from the snip.  The
-callback is invoked with the actual snip object, the @racket[mouse-event%] and
-the X, Y position of the mouse in plot coordinates -- these are in the same
-coordinate system used by the renderers in the plot.  The X, Y values are
-@racket[#f] when the mouse is outside the plot area (for example, when the
-mouse is over the axis area).
-
-When a callback is installed, the default zoom functionality of the plot snips
-is disabled.  This can be restored by calling
-@racket[set-mouse-event-callback] with a @racket[#f] argument.
-
-}
-
-@defmethod[(set-overlay-renderers [renderers (or/c (treeof renderer2d?) #f)]) any/c]{
-
-Set a list of renderers (or more generally, a tree of renderers) to be drawn
-on top of the existing plot.  This can be any combination of 2D renderers, but
-it will not be able to modify the axes or the dimensions of the plot area.
-Only one set of overlay renderers can be installed, calling this method a
-second time will replace the previous overlays.  Specifying @racket[#f] as the
-renderers will cause overlays to be disabled.
-
-}
-}
-
-@defthing[plot-mouse-event-callback/c contract? #:value (-> (is-a?/c snip%)
-                                                       (is-a?/c mouse-event%)
-                                                       (or/c real? #f)
-                                                       (or/c real? #f) any/c)]{
-A contract for callback functions passed to @racket[set-mouse-event-callback].
-}
-
-
-For example, if you evaluate the code below in DrRacket, the resulting plot
-will show a vertical line tracking the mouse and the current plot position is
-shown on a label.  This is achieved by adding a mouse callback to the plot
-snip returned by @racket[plot-snip].  When invoked, the mouse callback will
-add a @racket[vrule] at the current X position and a @racket[point-label] at
-the current value of the plotted function.
+If the code below is evaluated in DrRacket, the resulting plot will show a
+vertical line tracking the mouse and the current plot position is shown on a
+label.  This is achieved by adding a mouse callback to the plot snip returned
+by @racket[plot-snip].  When the mouse callback is invoked, it will add a
+@racket[vrule] at the current X position and a @racket[point-label] at the
+current value of the plotted function.
 
 @racketblock[
 (require plot)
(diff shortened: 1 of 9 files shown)


alex-hhh:
There is no documentation added for this functionality, not sure where to add it.

bennn:
I guess there are 2 choices for where to add documentation:

1. Make a new file like `plot-doc/plot/scribblings/overlay2d.scrbl`, 
2. Add a new section to the bottom of `plot-doc/plot/scribblings/renderer2d.scrbl`

My preference is `#2`, mostly because I like to use `Ctrl-f` to search single-page documents.

alex-hhh:
I have added documentation about this functionality, however I have run into some problems and I might need some help:

* The new methods are documented as if they have contracts attached, but they don't.  I would like to add contracts to the code, but I'm not sure how to do that without adding contracts to the entire `2d-plot-snip%` class
* the `plot-mouse-event-callback/c` mentioned in the documentation is not output as a hyperlink in the resulting HTML -- I suspect this is because there is no such contract defined in the code, and the scribble document imports the actual plot library, but I am not sure.
* the example code snipped given will print its output a plot image, but in a static document this is not very useful, but I'm not sure what `@` construct to use to get nicely formatted Racket code without evaluating it.
* related to the point above, is there a place where things like `@racket`, `@defclass`, `@defmethod`, etc are documented?  I could not find anything and my only scribble experience is trying to write documentation for these plot changes.

bennn:
> * The new methods are documented as if they have contracts attached, but they don't.  ....

It's ok to write a `class/c` contract that only names a few methods. Example: <http://pasterack.org/pastes/18339>

(I think this is exactly the right thing to do here, because it means method calls within the class will not get checked against the contract)

> * the `plot-mouse-event-callback/c` mentioned in the documentation is not output as a hyperlink in the resulting HTML ....

There needs to be a `@defproc[(plot-mouse-event-callback/c ...) ...]` for the hyperlink to go to.

> * the example code snipped given will print its output a plot image, but in a static document this is not very useful, but I'm not sure what `@` construct to use to get nicely formatted Racket code without evaluating it.

For formatting code, there's `racketblock`. But I like what you have currently, with `examples`

> * related to the point above, is there a place where things like `@racket`, `@defclass`, `@defmethod`, etc are documented?

They're hidden in the scribble docs: <http://docs.racket-lang.org/scribble/manual.html>

The `@` isn't part of the name, so that's why searching for `@racket` doesn't give any results. A good substitute is `racket M:scribble`, which searches for identifiers that start with Racket from modules whose names start with (or contain?) `scribble`.

- - -

You might have figured this out already, but Scribble has 2 modes --- "text mode" and "racket mode" --- and the `@`-sign switches between them:
- `@(f x)` is the same as `(f x)` in a Racket program
- `@f[x]` is the same as `@(f x)`, but preferred
- `@f{x}` is similar to `(f "x")` --- more precisely the stuff between the `{}` gets evaluated in "text mode" and `f` gets invoked on the result
- and `@x` or `@|x|` gets the "racket mode" value of the identifier `x`

Also, just in case, I think the fastest way to render the documentation is to run `raco setup plot`

plot-doc/plot/scribblings/renderer2d.scrbl
```diff
@@ -890,3 +890,99 @@ Returns a renderer that draws a labeled point on a polar function's graph.
...
+
+@itemlist[
+
+@item{@(racket hrule) and @(racket vrule) renderers can be used to draw
+horizontal and vertical lines that track the mouse position}
+
+@item{The @(racket rectangles) renderer can be used to highlight a region on
+the plot.  You can also specify @racket[-inf.0] and @racket[+inf.0] as limits
+for one of the dimensions, to extend the rectangle all the way to the edge of
+the plot in that direction.  You will need to specify a non zero value for the
```
bennn: suggestion: delete these two sentences ("You can .... You will ....") and skip to the example

bennn:
One more question: why does the `zoom-or-unzoom` handler use the `area-bounds->plot-bounds` function, but the `user-mouse-event-handler` use `(send area dc->plot ....)` ?

(I tried replacing `dc->plot` with `area-bounds->plot-bounds`, but that didn't just work.)

alex-hhh:
`area-bounds->plot-bounds` works on a rectangle (2 points), but `user-mouse-event-handler` only needs to transform one point.  You can actually use `area-bounds->plot-bounds`, but that would be inefficient:

```racket
(define (user-mouse-event-handler dc x y editorx editory evt)
      (define mouse-x (- (send evt get-x) x))
      (define mouse-y (- (send evt get-y) y))
      (if (rect-contains? area-bounds-rect (vector mouse-x mouse-y))
          (match-let (((vector (ivl px px1) (ivl py py1))
                       (area-bounds->plot-bounds
                        (vector (ivl mouse-x mouse-x) (ivl mouse-y mouse-y)))))
            (mouse-event-callback this evt px py))
          (mouse-event-callback this evt #f #f)))
```

The `area` object is needed in the `2d-plot-snip%` object because `draw-overlay-renderers` also uses `plot->dc`.  I will push some changes to remove the `area-bounds-rect` and `area-bounds->plot-bounds` fields from `2d-plot-snip%`, as they are redundant and the functions involved already take too many parameters (in my opinion anyway :) )

alex-hhh:
>>*  the plot-mouse-event-callback/c mentioned in the documentation is not output as a hyperlink in the resulting HTML ....
>
> There needs to be a `@defproc[(plot-mouse-event-callback/c ...) ...]` for the hyperlink to go to.

`plot-mouse-event-callback/c`  is a contract, not a procedure. All other contracts are defined using `@defthing`.

Thanks for the intro on Scribble -- indeed I used the search functionality to find things like `@defproc` and I could not, so I assumed it was not documented...

bennn:
> > There needs to be a `@defproc[(plot-mouse-event-callback/c ...) ...]` for the hyperlink to go to.
>
> plot-mouse-event-callback/c is a contract, not a procedure. All other contracts are defined using @defthing

Okay I was completely wrong about this: (1) you're right, `defthing` is better here, and (2) there's no link because there is no `for-label` binding for the contract.

Here's a smaller example of a document with some broken links:  [test.zip](https://github.com/racket/plot/files/1742612/test.zip)

To add a binding for this & the snip methods, they need to be exported from a (public) plot module and imported `for-label` in `plot-doc/plot/scribblings/common.rkt`.

I think a good way to do this is to make a new file `plot-gui-lib/plot/snip.rkt` that exports the snip and the contract and move the current documentation for overlays to a new file `plot-doc/plot/scribblings/snip2d.scrbl`. (I don't like multi-page documentation, but having a page for snip2d seems to fit with plots current docs.)

alex-hhh:
I have exported the `plot-mouse-event-callback/c' and also `2d-plot-snip+c%` from the main plot package -- it seems like a good idea to make these available to clients of the plot library.  Once I did that, the links were correctly generated.  I did not move the plot snip in a separate file, as this was not needed -- let me know if you want them separated into their own file.

plot-gui-lib/plot/main.rkt
```diff
@@ -2,12 +2,15 @@
 
 (require plot/no-gui
          "private/gui/plot2d.rkt"
-         "private/gui/plot3d.rkt")
+         "private/gui/plot3d.rkt"
+         "private/gui/snip2d.rkt")
```
bennn: I don't think this is a good idea, because of this code in `plot-gui-lib/plot/private/gui/lazy-snip-untyped.rkt`: 

```
;; Require lazily, in case someone wants to just (require plot) in a headless setup
(lazy-require ["snip2d.rkt" (make-2d-plot-snip)]
              ["snip3d.rkt" (make-3d-plot-snip)]
              ["gui.rkt" (make-snip-frame)])
```

Adding this require in `main.rkt`, I think, defeats the point of the lazy require. (There might be some programs that `(require plot)` that will fail because they're now forced to load the gui.)

bennn: ^ but this is a guess, and I'm not sure how to test the guess

alex-hhh: I think you are right.  The intent is that the GUI libraries are only initialized when a plot procedure is called. Simply adding `(require plot)` to a program should not cause these to be initialized.

I'm not sure how to solve this, though.  Can you `lazy-require` a contract and a class declaration?  Or should I just provide these `for-label` to make the doc links, but not make them available to the user?

bennn: This is why I suggested making a new module `plot/snip.rkt` that requires & provides the class & contract. Then users & the documentation (`common.rkt`) can require `plot/snip`, and programs that require `plot` don't change.

It might work to change the `require` in `plot-gui-lib/plot/main.rkt` to a lazy require.

And this might be a non-issue --- when I run an empty file `#lang racket/base (require plot)`, it seems to load the gui, so maybe the lazy require isn't helping ¯\_(ツ)_/¯

plot-gui-lib/plot/private/gui/snip2d.rkt
```diff
@@ -243,9 +356,18 @@
       (super resize w h))
     ))
 
+(define plot-mouse-event-callback/c (-> (is-a?/c snip%) (is-a?/c mouse-event%)
+                                        (or/c real? #f) (or/c real? #f) any/c))
+
+(define/contract 2d-plot-snip+c%
```
bennn: Using `define/contract` gives worse error messages than `contract-out` --- if a user gives a bad argument to a method, the error message will blame `plot-gui-lib/plot/private/gui/snip2d.rkt`

Here's a similar example, the outer module should get blamed, but the `define/contract` contract will blame the submodule `M`:  http://pasterack.org/pastes/35293

Suggestion:
- provide `2d-plot-ship%` and `make-2d-plot-snip` with `contract-out`
- then there's no need to define or document `2d-plot-snip+c%`, so remove it

(If you want `plot` to be able to make a `2d-plot-snip%` without a contract, then just move the `(contract-out [2d-plot-snip% ....])` to `plot-gui-lib/plot/snip.rkt`)

alex-hhh: I need some more guidance on your two suggestions.

>  *  provide 2d-plot-ship% and make-2d-plot-snip with contract-out

I'm not sure how to export `make-2d-plot-snip` with a contract.  First, it takes 7 arguments and I'm not sure what their contracts are, and `(-> any/c any/c ... 2d-plot-snip+c%)` would not be very elegant.  Second, `make-2d-plot-snip` creates a `2d-plot-snip%` instance and returns it.  Will `contract-out` attach the `2d-plot-snip+c%` class contract to an object?  I tried it out, see the diff at https://gist.github.com/alex-hhh/c8b9c9bb7de29a4c993cd99147fc9809, but this gives out an error when I try to create a plot:

```
instantiate: contract violation
  expected: class?
  given: 2d-plot-snip+c%
```

>  *  then there's no need to define or document 2d-plot-snip+c%, so remove it

We need to document the `set-mouse-event-callback` and `set-overlay-renderers` and the scribble documentation requires a class to wrap them (although my only experience with scribble is this pull request).   I could just use `snip%` , but that would be confusing in my opinion.  Or do you want to export `2d-plot-snip%` and use that symbol for the `@defclass`?

bennn: The contract for `make-2d-plot-snip` should be `(unconstrained-domain-> (instanceof/c (class/c ....)))` with the `class/c` contract for `2d-plot-snip%`.

(I just remembered `unconstrained-domain->` this morning. If that didn't exist, then I'd suggest the 7 `any/c`s --- because the point of the contract is to wrap the objects this function returns, not to check its arguments [which all come from the plot library])

For documentation, I just want to export `2d-plot-snip%`.

alex-hhh:
The `unconstrained-domain->` and `instanceof/c` worked, thanks for the help.   Can you please review the latest commit.

plot-gui-lib/plot/private/gui/snip2d.rkt
```diff
@@ -15,9 +15,18 @@
          "worker-thread.rkt"
          "snip.rkt")
 
-(provide make-2d-plot-snip
-         plot-mouse-event-callback/c
-         2d-plot-snip+c%)
+(define plot-mouse-event-callback/c
+  (-> (is-a?/c snip%) (is-a?/c mouse-event%) (or/c real? #f) (or/c real? #f) any/c))
```
bennn: Should `(is-a?/c snip%)` be `(is-a?/c 2d-plot-snip+c%)` ?

plot-gui-lib/plot/snip.rkt
```diff
@@ -4,4 +4,4 @@
 
 (provide
  plot-mouse-event-callback/c
- 2d-plot-snip+c%)
+ 2d-plot-snip%)
```
bennn: I think this should be `(contract-out [2d-plot-snip% 2d-plot-snip+c%])`

plot-gui-lib/plot/private/gui/snip2d.rkt
```diff
@@ -15,9 +15,18 @@
...
+(define 2d-plot-snip+c%
```
bennn: Since this identifier is just the contract, the name `2d-plot-snip%/c` would be more idiomatic*.

`*` More like the `class/c` identifier in `https://github.com/racket/draw/blob/master/draw-lib/racket/draw/private/contract.rkt`

stamourv:
I like the idea, and I don't think a review on my part would add much.

plot-gui-lib/plot/private/gui/snip2d.rkt
```diff
@@ -33,25 +48,41 @@
...
+
     (define left-click-x 0)
     (define left-click-y 0)
     (define left-drag-x 0)
     (define left-drag-y 0)
-    
+
     (define plot-bounds-rects empty)
+
+    (define (area-bounds->plot-bounds rect)
```
evdubs: Looks like this was moved from plot2d.rkt, but the `assert`ions were removed? Was this intended?

evdubs:
Looks good to me. I just had a question regarding the move of `area-bounds->plot-bounds` and the removal of some `assert` calls, but it's more of a curiosity than a concern.

plot-gui-lib/plot/private/gui/snip2d.rkt
alex-hhh: The `assert` is a typed racket thing, this is non typed code.  I don't know what the equivalent non-typed variant is and I'm not sure it would be valuable in this case.

plot-gui-lib/plot/private/gui/snip2d.rkt
alex-hhh: `2d-plot-snip+c%` already depends on `plot-mouse-event-callback/c` and I get a compilation error if I define it like that.  I tried using `letrec`, but that didn't work either:

```racket
(define-values
  (2d-plot-snip%/c plot-mouse-event-callback/c)
  (letrec ([plot-mouse-event-callback/c
            (-> (is-a?/c 2d-plot-snip%/c) (is-a?/c mouse-event%)
                (or/c real? #f) (or/c real? #f)
                any/c)]
           [2d-plot-snip%/c
            (class/c
             (set-mouse-event-callback (->m (or/c plot-mouse-event-callback/c #f) any/c))
             (set-overlay-renderers (->m (or/c (treeof renderer2d?) #f) any/c)))])
    (values 2d-plot-snip%/c plot-mouse-event-callback/c)))
```

If I try to use it I get:

```
2d-plot-snip%/c: undefined;
 cannot use before initialization
```

plot-gui-lib/plot/private/gui/snip2d.rkt
bennn: The asserts were checking that `area` and the bounds in `rect` were not false. A non-typed version could be `(unless area (error ....))`.

Removing these asserts looks safe to me, and the function seems to error anyway when I call it with a bad rectangle.

bennn:
@alex-hhh do you mind if I merge this as 1 big commit (instead of 11 commits) ?

alex-hhh:
@bennn , yes merging this as one commit will be better and the package will have a simpler history.  

Are there any outstanding tasks for me?

bennn:
> Are there any outstanding tasks for me?

I can't think of any :)

bennn:
Merged! https://github.com/racket/plot/commit/d363100c790dbdf8859b4412a80f993e4dc036eb

Issue #39, opened 2018-02-22
plot-frame's frame is too small


LiberalArtist:
I recently tried using `plot-frame` for the first time, and (at least on Mac OS) the resulting frame cuts off the plot half-way through the x-axis labels. Resizing the window causes the plot to be resized to fit correctly within the frame. 

Here are a minimal example and a screenshot:
```racket
#lang racket

(require plot)

(send (plot-frame (points '((2010 39.4)
                            (2010 93.8)
                            (2011 40.3)
                            (2011 79.9)
                            (2012 44.0)
                            (2012 85.5)))
                  #:x-label "Year"
                  #:y-label "Temperature [F]")
      show
      #t)
```
<img width="420" alt="plot-frame" src="https://user-images.githubusercontent.com/10700996/36554829-ea04b4ea-17c5-11e8-8498-dc9715cfc8d2.png">

alex-hhh:
Actually, the snip is too big :-) I proposed a fix for this.  The problem is present in Racket 6.11 as well...

bennn:
https://github.com/racket/plot/commit/b1a27914d88af828fb68105820808714e9a41127

PR #40, opened 2018-02-24
Construct the plot snip with the right size for plot-frame #39

alex-hhh:
The construction of the snip is delayed until its exact size is known by the
snip-canvas% that will hold it.

### DIFF
--- a/plot-gui-lib/plot/private/gui/gui.rkt
+++ b/plot-gui-lib/plot/private/gui/gui.rkt
@@ -16,9 +16,8 @@
     
     (super-new)))
 
-(define (make-snip-frame snip width height label)
-  (define (make-snip w h) snip)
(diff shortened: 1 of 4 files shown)


alex-hhh:
I should note that this could have been also fixed by resizing the snip inside the `make-snip` of the `plot-frame` call with `(send snip resize w h)`, but this would result in the plot bitmap being constructed with the wrong size, than constructed again with the right size (i.e. the plot is drawn twice).

bennn:
LGTM

alex-hhh:
If this fix is OK, can it be merged into master?

bennn:
merged https://github.com/racket/plot/commit/b1a27914d88af828fb68105820808714e9a41127

PR #41, opened 2018-03-01
Add the ability to not call into collapse-ticks

evdubs:
This is most helpful when trying to just have tick labels on the far y axis while keeping the ticks on the near y axis and allowing the grid to still be drawn. Here is the code that can be used for testing

```
#lang racket

(require plot)

(collapse-ticks? #t)

(parameterize ([plot-width    150]
                 [plot-height   150]
                 [plot-x-label  #f]
                 [plot-y-label  #f]
                 [plot-y-ticks (ticks (linear-ticks-layout) no-ticks-format)]
                 [plot-y-far-ticks (linear-ticks)])
    (plot (function sin (- pi) pi)))
```

By changing `collapse-ticks?` to `#f`, we can see all of the ticks on the near y axis and the `tick-grid` will also be displayed.

### DIFF
--- a/plot-lib/plot/private/common/parameters.rkt
+++ b/plot-lib/plot/private/common/parameters.rkt
@@ -83,6 +83,8 @@
 (defparam plot-y-far-axis? Boolean #t)
 (defparam plot-z-far-axis? Boolean #t)
 
+(defparam collapse-ticks? Boolean #t)
+
 (defparam2 plot-x-tick-label-angle angle Real Real 0 (rational 'plot-x-tick-label-angle))
 (defparam2 plot-y-tick-label-angle angle Real Real 0 (rational 'plot-y-tick-label-angle))
 (defparam2 plot-x-far-tick-label-angle angle Real Real 0 (rational 'plot-x-far-tick-label-angle))
--- a/plot-lib/plot/private/plot2d/plot-area.rkt
+++ b/plot-lib/plot/private/plot2d/plot-area.rkt
@@ -262,21 +262,25 @@
     (: x-far-ticks (Listof tick))
     (: y-far-ticks (Listof tick))
     (define x-ticks
-      (collapse-ticks (filter (λ ([t : tick]) (<= x-min (pre-tick-value t) x-max))
-                              (map tick-inexact->exact rx-ticks))
-                      (x-tick-near? y-min)))
+      (let ([exact-ticks-within-bounds (filter (λ ([t : tick]) (<= x-min (pre-tick-value t) x-max))
+                                               (map tick-inexact->exact rx-ticks))])
+        (if (collapse-ticks?) (collapse-ticks exact-ticks-within-bounds (x-tick-near? y-min))
+            exact-ticks-within-bounds)))
     (define x-far-ticks
-      (collapse-ticks (filter (λ ([t : tick]) (<= x-min (pre-tick-value t) x-max))
-                              (map tick-inexact->exact rx-far-ticks))
-                      (x-tick-near? y-max)))
+      (let ([exact-ticks-within-bounds (filter (λ ([t : tick]) (<= x-min (pre-tick-value t) x-max))
+                                               (map tick-inexact->exact rx-far-ticks))])
+        (if (collapse-ticks?) (collapse-ticks exact-ticks-within-bounds (x-tick-near? y-max))
+            exact-ticks-within-bounds)))
     (define y-ticks
-      (collapse-ticks (filter (λ ([t : tick]) (<= y-min (pre-tick-value t) y-max))
-                              (map tick-inexact->exact ry-ticks))
-                      (y-tick-near? x-min)))
+      (let ([exact-ticks-within-bounds (filter (λ ([t : tick]) (<= y-min (pre-tick-value t) y-max))
+                                               (map tick-inexact->exact ry-ticks))])
+        (if (collapse-ticks?) (collapse-ticks exact-ticks-within-bounds (y-tick-near? x-min))
+            exact-ticks-within-bounds)))
     (define y-far-ticks
-      (collapse-ticks (filter (λ ([t : tick]) (<= y-min (pre-tick-value t) y-max))
-                              (map tick-inexact->exact ry-far-ticks))
-                      (y-tick-near? x-max)))
+      (let ([exact-ticks-within-bounds (filter (λ ([t : tick]) (<= y-min (pre-tick-value t) y-max))
+                                               (map tick-inexact->exact ry-far-ticks))])
+        (if (collapse-ticks?) (collapse-ticks exact-ticks-within-bounds (y-tick-near? x-max))
+            exact-ticks-within-bounds)))
 
     ;; ===============================================================================================
     ;; Tick and label parameters, and fixpoint margin computation
--- a/plot-lib/plot/private/plot3d/plot-area.rkt
+++ b/plot-lib/plot/private/plot3d/plot-area.rkt
@@ -470,33 +470,39 @@
     (: y-ticks (Listof tick))
     (: z-ticks (Listof tick))
     (define x-ticks
-      (collapse-ticks (filter (λ ([t : tick]) (<= x-min (pre-tick-value t) x-max))
-                              (map tick-inexact->exact rx-ticks))
-                      (x-ticks-near? x-axis-y)))
+      (let ([exact-ticks-within-bounds (filter (λ ([t : tick]) (<= x-min (pre-tick-value t) x-max))
+                                               (map tick-inexact->exact rx-ticks))])
+        (if (collapse-ticks?) (collapse-ticks exact-ticks-within-bounds (x-ticks-near? x-axis-y))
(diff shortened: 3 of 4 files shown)


evdubs:
I'll need to add some documentation for this feature.

evdubs:
I would like to know if there's either a way to display the near y axis ticks that I haven't figured out or if there's a more appropriate fix.

alex-hhh:
I have only spend a small amount of time on this, and I didn't have time to figure out what `collapse-ticks` actually does (i.e. why are the ticks removed). I played with some variations of your example and I was surprised by the results...

My preference would be to find out if there is actually a bug in `collapse-ticks` and fix that, instead of bypassing the call altogether.  Alternatively, if you believe that `collapse-ticks` actually works correctly, it might be worth explaining that in this pull request.

BTW, I'm not a maintainer, just a user of this package, the maintainers might have different opinions on this.

bennn:
I don't see a way to get unlabeled near-axis ticks and labeled far-axis ticks with the current plot.

If there was a way to tell plot "use the right side as the near axis, not the left side", would that solve your problem? (I'm not sure how to implement that, but it seems like a clearer fix.)

evdubs:
@bennn that would be a preferable solution. I only took this approach because it seemed to work and was relatively straight forward.

evdubs:
@bennn do you think a change to `same-label?` like @alex-hhh proposed is a better solution?

bennn:
Yes, if we have a parameter to make the new behavior opt-in (to avoid changing existing plots). Something like this:

```
(define (same-label? t1 t2)
  (if (or (collapse-empty-ticks?)
          (non-empty-string? (tick-label t1)))
    (string=? (tick-label t1) (tick-label t2))
    #false))
```
This is better to me because it's a smaller change.

But I'd still prefer to have a parameter that makes the left side be the far axis.

evdubs:
@bennn the parameter should already be opt-in, it just has a larger scope if you opt in rather than just taking effect within the same-label? check. Or am I misunderstanding something?

evdubs:
@bennn I will take a look to see what the code change looks like to switch axes for being near/far.

bennn:
I think we understand each other
- `collapse-ticks?` is opt-in
- just changing `same-label?` would not be opt-in
- changing `same-label?` but adding a parameter would be opt-in

And to be clear I don't feel very strongly that `collapse-ticks?` is worse than changing `same-label?`. The most important thing is that the docs (for whatever this PR ends up doing) help other people find this solution.

alex-hhh:
Just to add my two cents to this conversation...

I think the current behavior of `collapse-ticks?` contains a bug.  If I read the code correctly, it should collapse ticks that are close to each other compared to the plot area, or whose printed representation is the same (e.g when using 1 digit precision for the numbers).   With the current implementation, however, it will consider two ticks to be the same if they have no label, regardless of whether they are actually close or not.

With the current behavior, `no-ticks-format` is useless, as all the ticks will be collapsed into one -- might as well just generate one tick in the middle of the interval.

bennn:
> I think the current behavior of collapse-ticks? contains a bug. ....
> With the current behavior, no-ticks-format is useless, ....

Whether or not it's a bug, I think we should keep the current behavior as the default --- so if a plot user upgrades to the next version of Racket, their old plots still look the same.

bennn:
ah! I just ran into a problem with `collapse-ticks`. Here's code for my plot:
```
#lang racket
(require plot/no-gui plot/utils)

(define data
  '((#(#\A 1.0) #(#\B 1.0) #(#\C 1.0) #(#\D 1.0) #(#\E 1.0) #(#\F 1.0) #(#\G 1.0) #(#\H 1.0) #(#\I 1.0) #(#\J 1.0) #(#\K 1.0) #(#\L 1.0) #(#\M 1.0) #(#\N 1.0) #(#\O 1.0) #(#\P 1.0) #(#\Q 1.01) #(#\R 1.0) #(#\S 1.0))
    (#(#\A 1.02) #(#\B 1.0) #(#\C 1.15) #(#\D 272.7) #(#\E 1.07) #(#\F 1.01) #(#\G 1.12) #(#\H 1.02) #(#\I 1.02) #(#\J 1.06) #(#\K 1.13) #(#\L 1.01) #(#\M 1.09) #(#\N 1.06) #(#\O 1.09) #(#\P 0.98) #(#\Q 1.14) #(#\R 34.98) #(#\S 0.96))))

(define my-ticks
    (ticks (λ (ax-min ax-max)
             (define major-ticks
               (for/list ((i (in-range 0 (log ax-max 10))))
                 (pre-tick (expt 10 i) #true)))
             (define minor-ticks
               (for*/list ((i (in-range 0 (exact-ceiling (log ax-max 10))))
                           (j (in-range 2 10 2)))
                   (pre-tick (* (expt 10 i) j) #false)))
             (append minor-ticks major-ticks))
           (lambda (ax-min ax-max pre-ticks)
             (for/list ([pt (in-list pre-ticks)])
               (format "10~a" (integer->superscript (exact-floor (log (pre-tick-value pt) 10))))))))

(parameterize ([plot-x-ticks no-ticks]
               [plot-y-ticks my-ticks]
               [plot-y-transform log-transform])
  (plot-file
    (append
      (for*/list ((i (in-range 0 4))
                  (j (in-range 2 10 2)))
        (hrule (* (expt 10 i) j) #:color 0 #:style 'solid #:alpha 0.2))
      (for/list ([color (in-list '(3 4))]
                 [style (in-list '(solid fdiagonal-hatch))]
                 [x-min (in-naturals 0)])
        (discrete-histogram
          (list-ref data x-min)
          #:x-min x-min
          #:skip 3
          #:label #f
          #:style style
          #:color (->brush-color color)
          #:line-color (->pen-color color)
          #:add-ticks? #true)))
    "sample.png"
    'png
    #:y-min 0.5
    #:width 400
    #:height 200))
```

Here's what the plot looks like on the current `master` branch. This is bad because there are no y minor ticks:
ticks-before

Here's what the plot looks like if I remove all calls to `collapse-ticks`. This is bad because the x ticks are not collapsed:
ticks-after

And here's what the plot looks like if I call `collapse-ticks` and change `same-label?` to Alex's definition above --- it looks the same as the first picture:
ticks-same-label

- - -

@evdubs what do you think about changing this PR from 1 parameter to 6 parameters (`plot-collapse-x-near-ticks?` ... `plot-collapse-z-far-ticks?`)?

evdubs:
I can certainly do that. Do you think it's still worth investigating the whole "treat the far-x-axis as the near-x-axis" thing?

alex-hhh:
Hi @benn, your minor ticks are missing because they have the same label as the major ticks, and this is exactly what `collapse-ticks` will remove.  Since the minor ticks don't display a label anyway, you can define them as:

```racket
(define my-ticks
    (ticks (λ (ax-min ax-max)
             (define major-ticks
               (for/list ((i (in-range 0 (log ax-max 10))))
                 (pre-tick (expt 10 i) #true)))
             (define minor-ticks
               (for*/list ((i (in-range 0 (exact-ceiling (log ax-max 10))))
                           (j (in-range 2 10 2)))
                   (pre-tick (* (expt 10 i) j) #false)))
             (append minor-ticks major-ticks))
           (lambda (ax-min ax-max pre-ticks)
             (for/list ([pt (in-list pre-ticks)])
               (if (pre-tick-major? pt)
                   (format "10~a" (integer->superscript (exact-floor (log (pre-tick-value pt) 10))))
                   (format "~a" (pre-tick-value pt)))))))
```

And they will show up as you intended.  This is with the code on the master branch. 

The changes I proposed to `same-label?` would only handle the case of empty labels and would have no effect on the rest of the `collapse-ticks` behavior, so it is unsurprising that it didn't fix the problem.  With the `same-label?` change, you could have left the minor pre-ticks with an empty label, but without it, you have to generate a unique label, which is than discarded.

bennn:
@alex-hhh thanks for this! I didn't realize the `ticks-format` function got called for every tick.

I'm still thinking parameters to disable `collapse-ticks` will be useful. Do you have an opinion on that?

alex-hhh:
Hi @bennn,

I don't think that a `collapse-ticks?` parameter would be useful.  The examples provided in this pull request don't justify it -- it is just that disabling the `collapse-ticks` function appeared to solve the immediate problem in each case.  `collapse-ticks` is also an internal function of the plot package, and the purpose of a `collapse-ticks?` parameter would be difficult to explain to someone who is not familiar with these internals.

For the first example, I think, some `labels-on-far-axis?`, `labels-on-near-axis?` parameters would be best.  The purpose of these parameters would be easy to explain to someone who only cared about the visual layout of a plot, without having to understand the internals of the plot package.

For the second example, I think the documentation for `ticks` could be improved to mention that all ticks generated by the function need to have a distinct label, including the minor ticks, otherwise they will be collapsed.

I also think the `same-label?` function should be fixed.  In such a case, the minor ticks could have an empty string as a label and they would not be collapsed.  In the current situation, the minor ticks have to have a label generated which is than discarded.

evdubs:
I created PR #43 and I think that code change will be much more agreeable.

PR #42, opened 2018-03-07
add arrows and arrows3d to allow plot arrows directly

pbpf:
add arrows and arrows3d to allow plot arrows directly;
now,we can do:

#lang racket
(require plot)
(plot3d  (arrows3d (list (vector (vector 0 0 0)  (vector    1 1 1))))
             ))

### DIFF
--- a/plot-lib/plot/private/common/parameters.rkt
+++ b/plot-lib/plot/private/common/parameters.rkt
@@ -196,7 +196,9 @@
 (defparam vector-field-scale (U Real 'auto 'normalized) 'auto)
 (defparam2 vector-field-alpha Real Nonnegative-Real 1 (unit-ivl 'vector-field-alpha))
 (defparam vector-field3d-samples Positive-Integer 9)
-
+;;arrow
+(defparam arrow-head-size-scale Nonnegative 1)
+(defparam arrow-head-angle-scale Nonnegative 1)
 ;; Error bars
 
 (defparam2 error-bar-width Real Nonnegative-Real 6 (nonnegative-rational 'error-bar-width))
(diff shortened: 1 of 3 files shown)


alex-hhh:
Remaining work on this pull request:

* [ ] document the new functions (`arrows` and `arrows3d`) plus the parameters in the plot manual
* [ ] not sure if using scaling for the arrow size and angle is the best way (my preference would be to specify the size directly, but I need to try things out first).
* [ ] add the plot parameters to one of the parameter groups, otherwise the plot snips will not use these parameters, see 
* [ ] add an example/test case of using these parameters, for an example see https://github.com/racket/plot/blob/master/plot-test/plot/tests/PRs/46.rkt

Given that this is an old pull request, I will work on these tasks, as time allows.

alex-hhh:
Hi @pbpf , I understand this is an old pull request and I am happy to complete the work to merge it in, but before I do that, can you please explain what types of data needs to be plotted using arrows? The `vector-field` renderer can already be used for vector fields, where the arrows make most sense, but I cannot think of any other use case.

bdeket:
I don't know the original use case for this pull request, but at some point I did something similar to make plots for a course in mechanics:
https://bdeket.synology.me/mechanica/0702circlevector.gif
I could help with this, but what would be the right way to work on someone else's PR?

alex-hhh:
Thanks for providing an example and for wanting to work on this.  The simplest way is to start a new branch in your repository, merge this branch into yours and continue the work, creating a new pull request.   Last I checked, the build was failing for this branch (the pull request was created before automated builds were set up).  Let me know if you need help with anything.

alex-hhh:
This is now done as part of #67, closing this pull request.

PR #43, opened 2018-04-21
add new params to control tick labels

evdubs:
This uses a different approach to showing ticks but hiding labels using feedback from @bennn and @alex-hhh on PR #41 

This was tested with the following code:
```
#lang racket

(require plot)

(parameterize ([plot-width    150]
                 [plot-height   150]
                 [plot-x-label  #f]
                 [plot-y-label  #f]
                 [plot-y-tick-labels? #f]
                 [plot-y-far-tick-labels? #t])
    (plot (function sin (- pi) pi)))
```
This will produce:
sin-no-labels
I suppose some documentation will be nice here even though this seems much more straight forward.

### DIFF
--- a/plot-lib/plot/private/plot2d/plot-area.rkt
+++ b/plot-lib/plot/private/plot2d/plot-area.rkt
@@ -331,10 +331,14 @@
     ;; -----------------------------------------------------------------------------------------------
     ;; Tick label parameters
 
+    (: draw-x-tick-labels? Boolean)
+    (: draw-y-tick-labels? Boolean)
     (: draw-x-far-tick-labels? Boolean)
     (: draw-y-far-tick-labels? Boolean)
-    (define draw-x-far-tick-labels? (not (and (plot-x-axis?) (equal? x-ticks x-far-ticks))))
-    (define draw-y-far-tick-labels? (not (and (plot-y-axis?) (equal? y-ticks y-far-ticks))))
+    (define draw-x-tick-labels? (plot-x-tick-labels?))
+    (define draw-y-tick-labels? (plot-y-tick-labels?))
+    (define draw-x-far-tick-labels? (or (plot-x-far-tick-labels?) (not (and (plot-x-axis?) (equal? x-ticks x-far-ticks)))))
+    (define draw-y-far-tick-labels? (or (plot-y-far-tick-labels?) (not (and (plot-y-axis?) (equal? y-ticks y-far-ticks)))))
 
     (: x-tick-label-offset (Vectorof Real))
     (: y-tick-label-offset (Vectorof Real))
@@ -355,7 +359,7 @@
 
     (: get-x-tick-label-params (-> (Listof Label-Params)))
     (define (get-x-tick-label-params)
-      (if (and (plot-x-axis?) (plot-x-tick-labels?))
+      (if (and (plot-x-axis?) draw-x-tick-labels?)
           (get-tick-label-params x-ticks
                                  x-tick-label-offset
                                  (λ ([x : Real]) (x-tick-value->dc x))
@@ -365,7 +369,7 @@
 
     (: get-y-tick-label-params (-> (Listof Label-Params)))
     (define (get-y-tick-label-params)
-      (if (and (plot-y-axis?) (plot-y-tick-labels?))
+      (if (and (plot-y-axis?) draw-y-tick-labels?)
           (get-tick-label-params y-ticks
                                  y-tick-label-offset
                                  (λ ([y : Real]) (y-tick-value->dc y))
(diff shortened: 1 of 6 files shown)


evdubs:
These new params use a similar naming convention to the label angles.

evdubs:
Paging @bennn @alex-hhh

plot-lib/plot/private/plot2d/plot-area.rkt
```diff
@@ -375,7 +375,7 @@
 
     (: get-x-far-tick-label-params (-> (Listof Label-Params)))
     (define (get-x-far-tick-label-params)
-      (if (and (plot-x-far-axis?) draw-x-far-tick-labels?)
+      (if (or (and (plot-x-far-axis?) draw-x-far-tick-labels?) (and (plot-x-far-axis?) (plot-x-far-tick-labels?)))
```
alex-hhh: I think that `draw-x-far-tick-labels?` should be updated to include `plot-x-far-tick-labels?` in the test and this line should remain unchanged.  There are other methods which might be affected by the position of the plot labels, for example `max-x-far-tick-label-height` and `get-x-far-tick-label-params`.

plot-lib/plot/private/plot2d/plot-area.rkt
```diff
@@ -355,7 +355,7 @@
 
     (: get-x-tick-label-params (-> (Listof Label-Params)))
     (define (get-x-tick-label-params)
-      (if (plot-x-axis?)
+      (if (and (plot-x-axis?) (plot-x-tick-labels?))
```
alex-hhh: We might need a new variable, `draw-x-tick-labels?`, similar to `draw-x-far-tick-labels?` as the code seems to assume that they are always drawn (see `max-x-tick-label-height`, `get-x-tick-label-params`),  this line should be

```racket
(if (and (plot-x-axis?) draw-x-tick-labels?) ...)
```

... and the other places updated too.

evdubs:
Have any thoughts @bennn ? Do you agree with @alex-hhh ?

bennn:
Thanks for the ping. I had tried making a test case that would show the problems Alex is worried about, but no luck so far. Hoping to try again tonight/tomorrow.

Can you add documentation for the new parameters?

alex-hhh:
With the original sample code given in this pull request, try zooming - unzooming the plot....

alex-hhh:
plot-bug

Here is what it looks like.  The problem is that plot parameters are taken from one plot-area to construct another one, and the information of where the labels is drawn is lost, since the relevant methods `get-*-parameters` only look at that `draw-x-far-tick-labels?`

evdubs:
Thanks for providing examples with buggy behavior, @alex-hhh . I'll try to fix this up shortly.

evdubs:
The code has been updated based on feedback. There are a few things to note:

1. max-z-tick-label-diag and max-z-far-tick-label-diag are missing. I don't know why this is or whether I should add them.

2. The issue of the DrRacket plot interaction seems to be a weird parameter issue. If I hard code the following, then it will keep the far tick labels and not show the near tick labels 
```
(defparam plot-y-tick-labels? Boolean #f)
(defparam plot-y-far-tick-labels? Boolean #t)
``` 
Also, without the hardcoding, it will keep the far tick labels and not show near tick labels when you view the plot in a new window. I would like to know if you attempt to reproduce that.

3. I still need to add documentation, but I'd like to wait until we have an agreed upon solution.

evdubs:
Paging @alex-hhh @bennn - let me know if I should stop doing this.

alex-hhh:
Hi @evdubs , the parameters don't maintain their values across different threads, and the `plot-area%` object is cloned in a different thread.  To get around this, the plot package defines parameter groups and stores / retrieves these as needed.  The new parameters you added will need to be added to the corresponding parameter group.

Below is the diff that fixes the problem: I added the parameters to the ` plot-tick-labels` group and also updated the `Plot-Parameters` type.  I don't think it really matters to which group they are added, as only the `plot-parameters` "super group" is used by the code, as far as I can tell.

The rest of the changes look good to me.

```diff
 plot-lib/plot/private/common/parameter-groups.rkt | 14 +++++++++++---
 1 file changed, 11 insertions(+), 3 deletions(-)

diff --git a/plot-lib/plot/private/common/parameter-groups.rkt b/plot-lib/plot/private/common/parameter-groups.rkt
index 3c2ffad..7b965e5 100644
--- a/plot-lib/plot/private/common/parameter-groups.rkt
+++ b/plot-lib/plot/private/common/parameter-groups.rkt
@@ -19,7 +19,14 @@
      plot-y-tick-label-anchor
      plot-y-tick-label-angle
      plot-y-far-tick-label-anchor
-     plot-y-far-tick-label-angle))
+     plot-y-far-tick-label-angle
+     plot-x-tick-labels?
+     plot-y-tick-labels?
+     plot-z-tick-labels?
+     plot-x-far-tick-labels?
+     plot-y-far-tick-labels?
+     plot-z-far-tick-labels?
+     ))
   
   (define-parameter-group plot-appearance
     (plot-width
@@ -89,7 +96,7 @@
       Anchor
       Nonnegative-Real
       (List Boolean Boolean Boolean Boolean Boolean Boolean)
-      (List Anchor Real Anchor Real Anchor Real Anchor Real)
+      (List Anchor Real Anchor Real Anchor Real Anchor Real Boolean Boolean Boolean Boolean Boolean Boolean)
       Boolean
       Boolean)
      (List Positive-Integer Real Real Nonnegative-Real Boolean Boolean)
```

evdubs:
I updated `parameter-groups.rkt` and added some documentation. Here's a screenshot
tick-labels-params-doc

evdubs:
@bennn have any thoughts? I think this is ready for review now.

bennn:
Looks good!

(Sorry for taking so long.)

PR #45, opened 2018-08-03
Capture plot parameters in plot-frame and plot3d-frame

alex-hhh:
`plot-frame` and `plot3d-frame` delays the creation of the plot snip object until the size of the canvas is known.  This means that the snip creation code is executed outside any `parameterize` scope and the values of any plot parameters (like `plot-decorations?`) would be ignored unless they are saved using `plot-parameters` and restored using `parameterize/group`

The fix is based on a problem reported here: https://groups.google.com/forum/#!topic/racket-users/ta-245wAYco

The issue was introduced when #39 was fixed, since at that time the plot snip was no longer created withing the scope of the `parameterize` call.

Here is some example code to test this (previously the two plots would have decorations, even though `plot-decorations?` is `#f):

```racket
#lang racket
(require plot)

(plot-new-window? #t)

(parameterize ((plot-decorations? #f))
  (plot3d (surface3d (λ (x y) (* (cos x) (sin y)))
                     -3.0 3.0 -3.0 3.0)))

(parameterize ((plot-decorations? #f))
  (plot (function sin) #:x-min -3 #:x-max 3))
```

### DIFF
--- a/plot-gui-lib/plot/private/gui/plot2d.rkt
+++ b/plot-gui-lib/plot/private/gui/plot2d.rkt
@@ -114,13 +114,17 @@
     [(and y-min (not (rational? y-min)))  (fail/kw "#f or rational" '#:y-min y-min)]
     [(and y-max (not (rational? y-max)))  (fail/kw "#f or rational" '#:y-max y-max)]
     [else
+     ;; make-snip will be called in a separate thread, make sure the
+     ;; parameters have the correct values in that thread as well.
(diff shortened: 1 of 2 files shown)


alex-hhh:
Hi @bennn , can you please review this pull request or ask someone else in the Racket dev team to review it?

This fixes a bug in the plot package were all plot parameters are ignored when either `plot-new-frame?` is `#t` or `plot-frame` is used.

Thanks,
Alex.

Issue #46, opened 2018-09-01
far tick labels drawn when parameter is #false


bennn:
The docs for `plot-x-far-tick-labels?` etc. say:

> When any of these is #f, the corresponding labels for the ticks on the axis are not drawn. 

But the corresponding labels _are_ drawn if the `plot-x-ticks` parameter (etc.) is overridden.

Example:
```
#lang racket/base
(require plot)
(parameterize* ([plot-x-far-ticks (plot-x-ticks)]
                [plot-y-far-ticks (plot-y-ticks)]
                [plot-x-ticks no-ticks]
                [plot-y-ticks no-ticks]
                [plot-x-far-tick-labels? #false]
                [plot-y-far-tick-labels? #false])
  (plot-pict
    (function values)
    #:x-min 0 #:x-max 2
    #:y-min 0 #:y-max 2))
```

Output:

![plot-far-tick-labels](https://user-images.githubusercontent.com/1731829/44949589-bbe03200-ae02-11e8-94f9-749b841b1ff2.png)

I think the docs are correct and the code should change somehow, maybe to make `plot-x-tick-labels? ` a 3-valued parameter.

alex-hhh:
Hi @bennn , I created a pull request to address this issue, can you have a look at it? Thanks, Alex.

bennn:
https://github.com/racket/plot/commit/9823d1ab9ff9144a8a3114afc0af6a69a0fbc4b7

PR #47, opened 2018-09-14
Allow plot-{x,y,z}-far-tick-labels? to have three options (issue #46)

alex-hhh:
The plot-{x,y,z}-far-tick-labels? parameters can have three possible values:
'auto, #t or #f.

When a parameter is set to 'auto, the behavior is that far labels are present
only if the ticks on the far axis are different from the ticks on the near
axis (this was the plot behavior before this parameter was introduced).

When a parameter is set to #t or #f, the far labels are shown or not
respectively.  Note that labels will only be displayed if the axis ticks are
drawn on the far axis.


plot-lib/plot/private/plot2d/plot-area.rkt
```diff
@@ -337,8 +337,16 @@
     (: draw-y-far-tick-labels? Boolean)
     (define draw-x-tick-labels? (plot-x-tick-labels?))
     (define draw-y-tick-labels? (plot-y-tick-labels?))
-    (define draw-x-far-tick-labels? (or (plot-x-far-tick-labels?) (not (and (plot-x-axis?) (equal? x-ticks x-far-ticks)))))
-    (define draw-y-far-tick-labels? (or (plot-y-far-tick-labels?) (not (and (plot-y-axis?) (equal? y-ticks y-far-ticks)))))
+    (define draw-x-far-tick-labels?
+      (let ((flag (plot-x-far-tick-labels?)))
+        (if (eq? flag 'auto)
+            (not (and (plot-x-axis?) (equal? x-ticks x-far-ticks)))
+            flag)))
```
bennn: Maybe replace this code with a function call `(f (plot-x-far-tick-labels?) (plot-x-axis?) x-ticks x-far-ticks)` ?

I can't think of what to name `f` though, so maybe not.

plot-test/plot/tests/pr46-test.rkt
```diff
@@ -0,0 +1,60 @@
...
+;; Case 6: far ticks are the same as near ticks, switch to drawing labels on the
+;; far axis
+(printf "Same ticks on near and far axis, labels on far axis only~%")
+(parameterize ([plot-width    150]
+               [plot-height   150]
+               [plot-y-tick-labels? #f]
+               [plot-y-far-tick-labels? #t]
+               [plot-y-ticks (linear-ticks)]
+               [plot-y-far-ticks (linear-ticks)])
+  (plot (function sin (- pi) pi)))
```
bennn: Can you set the x-tick parameters (the same way as the y-tick params) in all these tests, and add a few 3d tests for the z-params

alex-hhh:
I have updated the examples to include 3d plots -- it looks like the default `'auto` behavior for 3d plots is different than for the 2D plots, so I just changed the text to print out the parameter values that are used in each case.

With regards to documentation, I am not sure that the 'auto parameter should be documented -- it is the default when the user does not explicitly ask for far axis labels, I think the user should only ever set this parameter to `#t` or `#f`.  What do you think?

bennn:
> With regards to documentation, I am not sure that the 'auto parameter should be documented -- it is the default when the user does not explicitly ask for far axis labels, I think the user should only ever set this parameter to #t or #f. What do you think?

Sounds good. Can you delete the `#:value #f` in the docs, as in:

```
(defparam plot-x-far-tick-labels? draw? boolean? #:value #f) 
```

plot-test/plot/tests/pr46-test.rkt
```diff
@@ -0,0 +1,134 @@
+#lang racket
```
bennn: Oh, can you move this file to `plot-test/plot/tests/PRs/46.rkt`

bennn:
thank you! https://github.com/racket/plot/commit/9823d1ab9ff9144a8a3114afc0af6a69a0fbc4b7

PR #50, opened 2019-02-06
Update snip2d bitmap to final quality in 2d-plot-snip%/resize

alex-hhh:
The resize method is called when the snip is resized in the pasteboard.  In
practical applications of this, this method is called once when the pasteboard
itself is resized.  As such, the start-update-thread is called with a #f
parameter, which will result in the plot image being rendered at normal
quality.

Previously to this change, the plot image would first be rendered in low
quality, than once again in high quality, resulting in an ugly artefact when
the frame containing a plot is resized.

Note that in this version, we don't heck if the update thread is running,
instead the update thread is started unconditionally (it will stop a
previously running thread), so the final image is calculated.

This will affect cases where a plot snip is inserted in a pasteboard and
resized using the mouse, since in that case resize is called continuously,
however (1) even at low quality rendering, most plots will not be able to keep
up with dynamic resizing and (2) I am not aware of any applications which use
this feature anyway.


alex-hhh:
Here is a before image, illustrating the problem:

resize-before

And here is an after image:

resize-after

alex-hhh:
Hi @bennn , can you please review this pull request or ask someone else in the Racket team to review it?  Thanks, Alex.

bennn:
This looks ok to me. Let's merge

PR #51, opened 2019-02-08
Use platform specific bitmaps for rendering plots, where applicable

alex-hhh:
The 2D and 3D plot commands from the "no-gui" library have been updated to use
`make-bitmap`, while the corresponding commands from the GUI library have been
now use `make-platform-bitmap`.

`make-platform-bitmap` will create a bitmap optimized for being displayed on
the screen, with the correct backing scale for the monitor -- the
`draw-bitmap/supersampling` always used a backing scale of 2 which resulted in
blurred text for non-HiDPI screens, also Windows uses a backing scale of 1.5
by default on HiDPI displays.

The non-gui plot commands were already using `make-bitmap` via a call to
`draw-bitmap` but `draw-bitmap` and `draw-bitmap/supersampling` have been
removed from "draw.rkt", as they were no longer used, and they were not
providing much abstraction anyway -- also `make-platform-bitmap` is only
available in the GUI library.

### DIFF
--- a/plot-gui-lib/plot/private/gui/plot2d.rkt
+++ b/plot-gui-lib/plot/private/gui/plot2d.rkt
@@ -1,6 +1,7 @@
 #lang typed/racket/base
 
 (require (only-in typed/mred/mred Snip% Frame%)
+         (only-in racket/gui/base make-screen-bitmap)
          typed/racket/draw typed/racket/class racket/match
          plot/utils
          plot/private/common/parameter-group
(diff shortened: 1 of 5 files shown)


alex-hhh:
Hi @bennn , can you please review this pull request or ask someone else in the Racket team to review it?  Thanks, Alex.

bennn:
Thanks for the reminder. I will look at both pull requests this week.

bennn:
Does this work for plots that use an alpha channel on Windows? I'm asking because I read this sentence in the docs ([link](https://docs.racket-lang.org/draw/overview.html#%28part._.Portability%29)):

> On Windows, however, a bitmap produced by make-platform-bitmap has no alpha channel, and it uses more constrained resources than one produced by make-bitmap (due to a system-wide, per-process GDI limit).

alex-hhh:
My code uses `make-screen-bitmap`, but I used platform bitmap by mistake in the comments.  However, you are right that the alpha is not "fully" honored.  I wrote some tests and the `plot-background-alpha` will be applied to the background color, but it will be mixed with white instead of being added to the alpha channel.  The resulting bitmap will than be drawn onto the `editor-canvas%` without an alpha channel.

Now, there are a few things that need clarifying:

* `plot-bitmap` is not affected by this, as it uses `make-bitmap` and this will have an alpha channel.

* this only applies to plots which show up in DrRacket or shown using `plot-frame` and will affect only people who use DrRacket with a different background and use the alpha channel, however, the plot will not look nice even when using a normal `make-bitmap`:

capture

I could however change the bitmap to use a `make-bitmap` with the backing scale used by the system -- on Windows, the scale is 1.5, but previously the plot package always used 2.0.

```racket
(define bm (make-bitmap width height #t #:backing-scale (or (get-display-backing-scale) 1.0)))
```

I am not sure which one is better, the documentation is not very clear on that the differences are between `make-bitmap`, `make-platform-bitmap` and `make-screen-bitmap` and my informal testing showed that there is no significant speed improvement from using `make-screen-bitmap` over `make-bitmap`.  What do you think?

bennn:
What about first drawing into a `make-platform-bitmap` and then copying to a `make-bitmap`?

If that doesn't work well, then I think it's best to use `make-bitmap` with the system backing scale.

mflatt:
Wouldn't drawing to `make-platform-bitmap` first preserve the alpha-channel problem?

Using `make-bitmap` seems fine.

Using `make-bitmap` on Windows and `make-screen-bitmap` elsewhere probably would produce the best results across platforms. Of course, that route suggests yet another function from `racket/draw`, but if the existing options seem confusing, adding another one won't make it clearer.

(Meanwhile, it appears that text-rendering changes on Mojave have eliminated the difference between `make-screen-bitmap` and a scaled `make-platform-bitmap` on my screen, but I have no idea whether Mojave adapts to different displays.)

alex-hhh:
@bennn I will update the code to just use `make-bitmap` on all platforms. On Windows, at least, there is no visible performance gain from using `make-screen-bitmap` or `make-platform-bitmap` (and I have some plots that do take a long time to draw).

Ideally, I would like the plot snips to draw directly to the canvas, as I believe the intermediate step of drawing to a bitmap than copying it to the canvas is somewhat inefficient (of course, I would have to copy the bitmap out of the canvas to avoid performance degradation on redraw).  But this is a project for another day.

alex-hhh:
HI @bennn , I changed the code to use `make-bitmap` with the current display backing scale.

PR #52, opened 2019-03-21
Add color map support for selecting colors for renderers

alex-hhh:
Plot renderers have a `#:color` argument which specifies the color to use for drawing, this can be a `color%` object, a RGB triplet, or an integer representing an index into a color palette.  This commit introduces a mechanism to select which color palette (called a color map) to use when integers are used for the `#:color` argument of the renderers.

```racket
(require plot)
(parameterize ([plot-pen-color-map 'set1]
               [line-width 3])
    (plot (list (function sin -5 5 #:color 0)
                (function cos -5 5 #:color 1))))
```

sample-plot

The following new API functions are introduced and updated:

* `plot-pen-color-map` and `plot-brush-color-map` are new parameters which specify the name of a color map to use for the `#:color` argument of renderer functions.

* the `->pen-color` and `->brush-color` functions have been updated to use the current color map for lookups. These functions are used internally to convert integers to a RGB triplet, but they are also available in `plot/utils`.

* The `plot-foreground` and `plot-background` parameters have been changed to "black" and "white" respectively, this was done to avoid drawing the axes and text in the first color of the pen color map and the background in the first color of the brush color map.

* `color-map-names` is a new function which returns the list of color map names as a list of symbols.

* `color-map-size` is a function which returns the number of colors in a color map.  Note that `->pen-color` and `->brush-color` will cycle the colors in the color map, so any integer argument is valid for these functions.

* `register-color-map` allows registering new color maps.

The last three functions are available in the `plot/utils` module, but not `plot`.  This is consistent with the availability of `->pen-color`.

The following color map names are also added, they are based on the matplotlib and ggplot2 color maps and use the same names: dark2, paired, pastel1, pastel2, set1, set2, set3, tab10, tab10n, tab20, tab20b and tab20c.

color-maps

The default plot colors are not changed with this commit, if the user does not specify a color map using `plot-pen-color-map` or `plot-brush-color-map`, the plot will use the same colors as before.

Some functions have been moved to different files, this was done to avoid circular dependencies.

# Some other notes

* There is a `racket-lang` discution about this feature [here](https://groups.google.com/forum/#!topic/racket-users/lGPYVc7s3ww)

* This pull request does not cover using contour colors for various 3d plots or changing the defaults for the color maps, these can be addressed in subsequent pull requests

* The color maps use list of three bytes as color definitions, this is consistent with what `->pen-color` returns, and has not changed -- it is not very convenient for using this feature outside the plot package, but it is consistent with how the plot package works.

* Once everyone is happy with the defined API, I will update the Scribble documentation and push the changes to this pull request.

The color map picture was generated using the code below, which also illustrates how to use `color-map-names` and `color-map-size`:

```racket
#lang racket
(require plot plot/utils pict racket/draw)

(define (pp-color-maps (width 400) (height 30))
  (define cm-names (sort (color-map-names)
                         (lambda (a b)
                           (string<=? (symbol->string a) (symbol->string b)))))
  (define cm-labels
    (for/list ([cm cm-names])
      (text (symbol->string cm) null 16)))
  (define cm-picts
    (for/list ([cm cm-names])
      (parameterize ([plot-pen-color-map cm])
        (define w (/ width (color-map-size cm)))
        (apply
         hc-append 0
         (for/list ([c (in-range (color-map-size cm))])
           (match-define (list r g b) (->pen-color c))
           (define color (make-object color% r g b))
           (filled-rectangle w height #:draw-border? #f #:color color))))))
  (define picts
    (let loop ([result '()]
               [labels cm-labels]
               [picts cm-picts])
      (if (null? labels)
          (reverse result)
          (loop (cons (car picts) (cons (car labels) result))
                (cdr labels)
                (cdr picts)))))
  (table 2 picts lc-superimpose cc-superimpose 15 3))
```

### DIFF
--- a/plot-doc/plot/scribblings/utils.scrbl
+++ b/plot-doc/plot/scribblings/utils.scrbl
@@ -190,11 +190,11 @@ Use @(racket ->pen-color) and @(racket ->brush-color) to convert integers.
 }
 
 @defproc[(->pen-color [c plot-color/c]) (list/c real? real? real?)]{
-Converts a @italic{line} color to an RGB triplet. This function interprets integer colors as darker and more saturated than @(racket ->brush-color) does.
 
-Non-integer colors are converted using @(racket ->color).
-Integer colors are chosen for good pairwise contrast, especially between neighbors.
-Integer colors repeat starting with @(racket 128).
+Convert a @italic{line} color to an RGB triplet. Integer colors are looked up
+in the current @racket[plot-pen-color-map], and non-integer colors are
+converted using @(racket ->color).  When the integer color is larger than the
+number of colors in the color map, it will wrap around.
 
 @examples[#:eval plot-eval
                  (equal? (->pen-color 0) (->pen-color 8))
@@ -204,12 +204,15 @@ Integer colors repeat starting with @(racket 128).
                         #:colors (map ->pen-color (build-list 8 values))))]
 }
 
-@defproc[(->brush-color [c plot-color/c]) (list/c real? real? real?)]{
-Converts a @italic{fill} color to an RGB triplet. This function interprets integer colors as lighter and less saturated than @(racket ->pen-color) does.
+The example above is using the internal color map, with
+@racket[plot-pen-color-map] set to @racket[#f].
 
-Non-integer colors are converted using @(racket ->color).
-Integer colors are chosen for good pairwise contrast, especially between neighbors.
-Integer colors repeat starting with @(racket 128).
+@defproc[(->brush-color [c plot-color/c]) (list/c real? real? real?)]{
+
+Convert a @italic{fill} color to an RGB triplet.  Integer colors are looked up
+in the current @racket[plot-brush-color-map] and non-integer colors are
+converted using @(racket ->color).  When the integer color is larger than the
+number of colors in the color map, it will wrap around.
 
 @examples[#:eval plot-eval
                  (equal? (->brush-color 0) (->brush-color 8))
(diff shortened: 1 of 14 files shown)


alex-hhh:
I have also updated the scribble documentation for the new functions.  This pull request now has all the changes I proposed.

alex-hhh:
Hi @bennn , can you please review this pull request or ask someone else in the Racket team to review it? Thanks, Alex.

bennn:
I can't review this until next weekend.

I think others are aware of this PR because of the mailing list discussion, but I'll advertise it on the Racket slack channel tomorrow & see if that catches anyone.

jeapostrophe:
I've reviewed the code and will merge it.

plot-doc/plot/scribblings/utils.scrbl
```diff
@@ -251,6 +258,28 @@ Integer brush styles repeat starting at @(racket 7).
...
+Return the number of colors in the color map @racket[name].  If @racket[name]
+is not a valid color map name, the function will signal an error.
+
+}
+
+@defproc[(register-color-map (name symbol?) (color-map (vectorof (list byte? byte? byte?)))) void]{
+
+Register a new color map @racket[name] with the colors being a vector of RGB
+triplets.  If a color map by that name already exists, it is replaced.
+
```
jeapostrophe: Can you add @history annotations on these new functions?

alex-hhh: Will do that in a separate pull request

Issue #48, opened 2018-11-01
Minor typo in polar->cartesian function & verification of longitude/latitude proper use.


encomer:
In the documentation of polar->cartesian function, currently it is writen:

"Converts 2D polar coordinates to 3D cartesian coordinates", but should say 2D.

Also, **please verify** the correctness of the following paragraph after the polar3d function documentation, concerning the ranges of latitudes and longitudes. 

"Currently, latitudes range from 0 to (* 2 pi), and longitudes from (* -1/2 pi) to (* 1/2 pi). These intervals may become optional arguments to polar3d in the future."

Thank you for the excellent plot module.

bennn:
> In the documentation of polar->cartesian function, currently it is writen:
> 
> "Converts 2D polar coordinates to 3D cartesian coordinates", but should say 2D.

Fixed in 13fce6908b68c4b2e4347efd402030c6e6b09d6d

> Also, please verify the correctness of the following paragraph after the polar3d function documentation, concerning the ranges of latitudes and longitudes.
>
> "Currently, latitudes range from 0 to (* 2 pi), and longitudes from (* -1/2 pi) to (* 1/2 pi). These intervals may become optional arguments to polar3d in the future."

Yes, the paragraph is still correct. Here is where the intervals are hard-coded:
https://github.com/racket/plot/blob/master/plot-lib/plot/private/plot3d/isosurface.rkt#L300

encomer:
Thank you @BourgondAries for directing me to the right repository.

And thanks @bennn for fixing the typo. **Still**, I would like to ask you concerning the names of latitudes and longitudes: are they refering to the correct concept? or do we need to swap them? (_I understand that the issue is closed_).

Thanks for all the great work that you are doing.

bennn:
Oh, I see. I'm not familiar with that use, but the docs seem self-consistent and their symbols for latitude/longitude seem to match the definitions here: https://en.wikipedia.org/wiki/Azimuthal_equidistant_projection

encomer:
Thank you @bennn for the reference. Still, please consider checking if in the **polar3d** documentation the term latitude should say longitude and viceversa. 
Extra references: 
[https://msdn.microsoft.com/en-us/library/aa578799.aspx](url)
[https://gisgeography.com/latitude-longitude-coordinates/](url)
_Note_: we can observe that the term latitude have the range (* -1/2 pi) to (* 1/2 pi).

bennn:
[...]

PR #56, opened 2019-08-15
match use of latitude and longitude, and their domain

bdeket:
This addresses issue #48 
to add my 2cts to the discussion:
latitude and longitude are used for navigation on a/our globe
latitude is from pole to pole: ie from 90°South to 90°North,
Longitude is from 180°West to 180°East
so latitude conforms with the (- (/ pi 2)) to (/ pi 2) and longitude goes from (- pi) to pi (or alternatively from 0 to (* 2 pi))

For polar coordinates you have one angle going trough 360° and one going trough 180°
in the implementation the first angle (theta) is going trough 360° and the second trough 180°
This is how it is usually done, with the understanding that the 360° usually goes around the equator (here hardwired to the xy-plane) and the second angle from pole to pole

The mix up in my opinion is because in math you usually specify first the 360° angle, then the 180° where in navigation you first say latitude, then longitude.

I also changed the symbols because rho is used for a radius (or a projected radius in azimuth projection) whereas theta and phi are more mainstream for the angles.

### DIFF
--- a/plot-doc/plot/scribblings/renderer3d.scrbl
+++ b/plot-doc/plot/scribblings/renderer3d.scrbl
@@ -161,20 +161,21 @@ Returns a renderer that plots a two-input, one-output function. For example,
           [#:alpha alpha (real-in 0 1) (surface-alpha)]
           [#:label label (or/c string? #f) #f]
           ) renderer3d?]{
-Returns a renderer that plots a function from latitude and longitude to radius.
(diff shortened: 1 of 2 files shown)


alex-hhh:
Hi @bennn , @jeapostrophe can you please have a look at this pull request?  The changes seem correct to me: navigation coordinates are opposite to the mathematical convention.

PR #57, opened 2019-09-27
Add #:scientific? option to linear-ticks-format and linear-ticks

gus-massa:
The idea is to be able to get an scale like

    8000    9000    10000    11000

instead of

    8000    9000     10^5    11000

using 

    (plot-x-ticks (linear-ticks #:scientific? #f))

Also, fix a few typos in the docs.


gus-massa:
Minor change of the version in the history note in docs. `7.4.0.9` -> `7.4.0.10`.

PR #59, opened 2020-06-15
Add #:invert? option to error-bars

lexi-lambda:
This makes it possible to add horizontal error bars to a horizontal histogram.

### DIFF
--- a/plot-doc/plot/scribblings/renderer2d.scrbl
+++ b/plot-doc/plot/scribblings/renderer2d.scrbl
@@ -150,6 +150,7 @@ An example of automatic scaling:
           [#:line-style line-style plot-pen-style/c (error-bar-line-style)]
           [#:width width (>=/c 0) (error-bar-width)]
           [#:alpha alpha (real-in 0 1) (error-bar-alpha)]
+          [#:invert? invert? boolean? #f]
           ) renderer2d?]{
 Returns a renderer that draws error bars.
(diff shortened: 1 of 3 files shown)


alex-hhh:
HI @lexi-lambda , I had a look at the changes and the look good, I would like to ask you to make two changes:

* pull from master into your branch, as I have configured a CI build, and I would like to see if this works for your branch
* can you please add a test case to https://github.com/racket/plot/tree/master/plot-test/plot/tests/PRs, currently there are no automated tests, so all I would like is a small example illustrating the new `#:invert?` option on the error bars

Thanks,
Alex.

alex-hhh:
I was a bit confused by the way this works, as I was expecting to simply have to set `#:inverted?` to `#t`, but it seems that one also has to swap the coordinates.  This, however seems consistent with the way `discrete-histogram` works...

```racket
#lang racket
(require plot)
(plot-width 300)
(plot-height 300)
(list
(plot (list (function sqr 1 7)
              (error-bars (list (vector 2 4 12)
                                (vector 4 16 20)
                                (vector 6 36 10))
                          #:invert? #f)))
;; NOTE: both the invert? parameter and the coordinates need to be inverted.
(plot (list (function sqr 1 7)
              (error-bars (list (vector 4 2 12)
                                (vector 16 4 20)
                                (vector 36 6 10))
                          #:invert? #t))))
```

pr59

PR #60, opened 2020-06-30
Setup CI for the plot package

alex-hhh:
This sets up a CI build for the plot package, scheduled to run for pushes and pull requests.  Currently it only runs for Racket version 7.7 and it does not run the tests.  Tests need are currently run manually and their output visually inspected.

It is inspired by the web-server CI build, but uses a different strategy for the build:

* minimal racket is installed
* a package catalog is setup for the packages in the plot repository
* this package catalog is added to the default racket catalogs
* the plot package is installed -- this will fetch packages form the local file system and build them, and will fetch all dependencies from the main racket catalog.

NOTE: there are several commits in this branch, as I was experimenting with different approaches, but I will squash the commits when merged to the master branch, so only one commit shows up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,5 +21,5 @@ jobs:
       - run: raco pkg config catalogs >> catalogs.txt
       - run: raco pkg config --set catalogs `cat catalogs.txt`
       - run: raco pkg config catalogs
-      # - run: sudo raco pkg install --batch --auto plot
+      - run: sudo raco pkg install --batch --auto plot
       # - run: raco test --drdr plot-test/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
           variant: ${{ matrix.racket-variant }}
           version: ${{ matrix.racket-version }}
       - run: racket -l- pkg/dirs-catalog --link catalog .
-      - run: CPATH=`pwd` echo "file://$CPATH/catalog" > catalogs.txt
+      - run: echo "file://`pwd`/catalog" > catalogs.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+          distribution: minimal
+          variant: ${{ matrix.racket-variant }}
+          version: ${{ matrix.racket-version }}
+      - run: racket -l- pkg/dirs-catalog --link catalog .
+      - run: echo "file://`pwd`/catalog" > catalogs.txt
+      - run: raco pkg config catalogs >> catalogs.txt
+      - run: raco pkg config --set catalogs `cat catalogs.txt`
+      - run: raco pkg config catalogs
+      - run: sudo raco pkg install --batch --auto plot
+      # - run: raco test --drdr plot-test/
```
samth, line 25: You might need `xvfb` here; see the use at https://github.com/racket/typed-racket/blob/master/.github/workflows/ci.yml#L29

.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
+on: [push, pull_request]
+name: CI
+jobs:
+  build:
+    name: "Build on Racket '${{ matrix.racket-version }}' (${{ matrix.racket-variant }})"
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        racket-version: ["7.7"]
```
samth, line 9: Probably good to test snapshots as well.

alex-hhh: I added the "current" as well as CS variants for Racket to the built, although, I am not convinced of their value to the build of plot -- "current" will be tested once the changes are merged to master anyway.  OTOH, we don't pay for the build VM time...

PR #55, opened 2019-08-14
Parametric surface3d

bdeket:
Adds a 3d renderer for paramteric surfaces

this is faster than the trick used for polar3d, where an isosurface is used, since the right values for the polygons can be directly computed

also exposes a renderer for polygons, similar to lines3d

PR #62, opened 2020-07-16
Implement new method for writing plot tests

alex-hhh:
### The Problem

Most of the tests in the `plot-test` package are written as Racket programs which output plot images to the DrRacket REPL.  When these tests are run automatically by "raco test" commands or [DrDr](http://drdr.racket-lang.org/), the only thing tested is that they don't report errors, any changes in the way plots are rendered will not be reported as test failures.

In addition to that, while the user can run these tests manually to inspect the plot images, they have no reference images to compare them to.  This means that they would only be able to detect the most obvious failures, not the subtle ones.

### The solution

This pull request introduces the a new way of writing plot tests where a sample plot is recorded and compared against the current rendering of the plot.  These types of tests have the ability to automatically detect changes in the way the plots are rendered.

This pull request only adds tests for changes in #5 and #59, since these were recently merged and had no tests, as well as #55 since this pull request used 3D plots, and I wanted to validate the method for 3D plots as well.

### The Implementation

Instead of comparing bitmap or vector images, plots are rendered using `plot/dc` or `plot3d/dc` to a special `record-dc%` object and the draw steps are written to a file. When running the tests, the new set of draw steps are generated and compared to the saved ones.

The `record-dc%` used for drawing is updated to report consistent values for text dimensions, regardless of the font -- this allows the data files to be used across different platforms which would generate slightly different plot images due to different fonts being available.

In addition to this, a sample PNG image is saved for each plot -- this allows the user to manually run the tests and having something to compare the result.

The verification steps are automated in `check-draw-steps` and `check-draw-steps3d` such that, when a verification fails, a new set of draw steps are written to file, as well as a new sample image.

The GitHub Actions `ci.yml` has been updated to upload these files as an artifact, so we have the data to analyze failed runs on GitHub Actions. The test will work fine without this feature, so they can be run on DrDr or the users computer.


rfindler:
Nice! I should change the bitmap tests in redex to follow this strategy.

samth:
Very nice.

alex-hhh:
The simplest way to make new RKTD files is to just remove the old ones and run the tests.  For new tests, start without a RKTD file and just run the test.

When a test fails (either because the draw steps have changed, or there is no draw steps file), it `check-draw-steps` will create new draw step files and sample images (these names start with "new-...").  These generated files can than be renamed onto the required file name and used in the tests.   This is why `generate-draw-steps` is not exported.

I plan to migrate the other tests to use this method, and based on that migration experience, I will probably refine the process a bit more.  Once I am done with the conversion, I will write up a "how to write plot tests" guide which can be used by people updating this package.

alex-hhh:
> Future thought: these tests and images might be a good basis for an extra plot gallery to go alongside the docs.

I am not sure that these images would be useful for that purpose -- a plot gallery would need to showcase the best that the plot package has to offer, but these tests would test the simplest use case, to keep the tests simple.

Creating a plot gallery, similar to the [Matplotlib Gallery](https://matplotlib.org/gallery.html) would be a nice to have, but without additional contributors, it will probably not happen anytime soon -- my plan for the plot package is to update the tests than start addressing the existing issues.  Given my limited time, this will keep me busy for a long time.

bennn:
ok sounds good

PR #69, opened 2020-09-01
vector-field auto scale

bdeket:
vector-field fails with 'division by zero' error when all vectors are horizontal or vertical. This PR fixes this (in the same way that it is addressed in vector-field3d)
```
(plot (vector-field (λ (x y) (list x 0))
                    -10 10 -10 10
                    #:scale 'auto))
```


plot-lib/plot/private/plot2d/point.rkt
```diff
@@ -126,8 +126,8 @@
                      [(? real?)  (map (λ ([mag : Real]) (* scale mag)) mags)]
                      ['normalized  (define box-size (min box-x-size box-y-size))
                                    (build-list (length dxs) (λ _ box-size))]
-                     ['auto  (define dx-max (apply max (map abs dxs)))
-                             (define dy-max (apply max (map abs dys)))
+                     ['auto  (define dx-max (real->double-flonum (apply max (map abs dxs))))
+                             (define dy-max (real->double-flonum (apply max (map abs dys))))
```
alex-hhh, line 130: This is a subtle fix and it took me a while to figure it out.  Can you add a comment in the code explaining why we use `real->double-flonum` here?  Something like:

> When all `dxs` or all `dys` are zero (vectors are aligned to the x or y axis), the calculation of `scale` below will raise a division by zero error.  If we convert these values to flonums, the result of such a division will be `+inf.0` and the correct scale will be calculated.

alex-hhh (commented):
Thanks for tracking down this defect.  I only have a single comment, relating to adding the code to explain the fix, but other than that, it looks good.

Issue #49, opened 2018-11-09
return legend as pict?


bennn:
Currently, plot draws legends on the plot area.

It would be nice if plotting functions could return the plot and legend as two picts that could be combined using a [pict combiner](http://docs.racket-lang.org/pict/Pict_Combiners.html) or [progressive pict](http://docs.racket-lang.org/ppict/index.html).

alex-hhh:
These are some notes for adding support to draw the plot legend outside the plot area.  Doing this is surprisingly difficult, as the notes below show.

### Define new legend anchor types

The location of the plot legend is controlled by the `plot-legend-anchor` parameter, which is of type `Anchor`, defined here: https://github.com/racket/plot/blob/4df34897acba74f689954188183e367c160911b8/plot-lib/plot/private/common/types.rkt#L15

We need to add `'outer-top-left`, `'outer-left`, `'outer-bottom-left`, `'outer-top-right`, `'outer-right` and `'outer-bottom-right` cases to this type.

**NOTE** unfortunately, this type is also used as an anchor to the labels on the plot and the new types would be invalid values for use as plot labels. Also the anchor type already contains `'auto`, which is used by the plot labels (meaning to place the label such that it fits inside the plot), but has no meaning for legend entries.  Perhaps we should separate the types and use `Anchor` for labels and a new type `LegendAnchor` for plot legends.

### Separate the legend size calculation from the drawing

https://github.com/racket/plot/blob/4df34897acba74f689954188183e367c160911b8/plot-lib/plot/private/common/plot-device.rkt#L613

The legend is actually drawn in `plot-device%/draw-legend`, this function both calculates the size of the legend and draws it.  It receives the legend entries, plus the plot area in plot coordinates (since, as of now, the legend is drawn inside the plot area).

We need at least a function which calculates the `legend-x-size` and `legend-y-size` separately.

Also this function will need to be updated to recognize the new legend anchors and adjust the position of the legend accordingly.

### Make room for the legend when it is outside

When the legend is outside the plot, space will need to be reserved for it. This is done here (there is an equivalent place for 3D plots): https://github.com/racket/plot/blob/4df34897acba74f689954188183e367c160911b8/plot-lib/plot/private/plot2d/plot-area.rkt#L554

The `get-param-vs/set-view->dc!` will need to be updated to add an entry for the legend, if the legend is outside.  The function returns a list of *entries* for `margin-fixpoint` to use.  Each entry is in the format:

```
(list label (vector x y) Anchor Angle)
[...]

PR #72, opened 2020-09-10
[Legend pt3] Make legend available to area at start

bdeket:
This build on #70 and #71 in order to implement #49.

Although all test pass, a little improvement can be made (changing 3 test cases), see [plot2d-utils#L73](https://github.com/racket/plot/blob/939390c7b3d5a858d7f1c5af2b246211bdbff83d/plot-lib/plot/private/no-gui/plot2d-utils.rkt#L73) and [plot3d-utils#L83](https://github.com/racket/plot/blob/939390c7b3d5a858d7f1c5af2b246211bdbff83d/plot-lib/plot/private/no-gui/plot3d-utils.rkt#L83)

### DIFF
--- a/plot-lib/plot/private/common/contract.rkt
+++ b/plot-lib/plot/private/common/contract.rkt
@@ -34,6 +34,8 @@
 (define anchor/c (one-of/c 'top-left    'top    'top-right
                            'left        'center 'right
                            'bottom-left 'bottom 'bottom-right))
+(define legend-anchor/c (or/c #f anchor/c
+                              (list/c (one-of/c 'inside 'outside) anchor/c)))
 
 (define color/c (or/c (list/c real? real? real?)
                       string? symbol?
--- a/plot-lib/plot/private/no-gui/plot2d-utils.rkt
+++ b/plot-lib/plot/private/no-gui/plot2d-utils.rkt
@@ -70,10 +70,11 @@
   
   (for ([rend  (in-list renderer-list)])
     (match-define (renderer2d rend-bounds-rect _bf _tf label-proc render-proc) rend)
+    ;; next step could be moved into (when render-proc, but this generates different step files)
+    (send area start-renderer (if rend-bounds-rect
+                                  (rect-inexact->exact rend-bounds-rect)
+                                  (unknown-rect 2)))
     (when render-proc
-      (send area start-renderer (if rend-bounds-rect
-                                    (rect-inexact->exact rend-bounds-rect)
-                                    (unknown-rect 2)))
       (render-proc area)))
   
   (send area end-renderers)
@@ -103,9 +104,8 @@
      (flatten-legend-entries
       (for*/list : (Listof (Treeof legend-entry))
         ([rend  (in-list renderer-list)]
-         [rect (in-value (let ([r (plot-element-bounds-rect rend)])
-                           (or (and r (clip r)) outer-rect)))]
          [label-proc (in-value (renderer2d-label rend))]
          #:when label-proc)
-        (label-proc rect)))]
+        ;; clipping not necessary, levels are always calculated on the complete bounds-rect
+        (label-proc outer-rect)))]
     [else '()]))
--- a/plot-lib/plot/private/no-gui/plot2d-utils.rkt
+++ b/plot-lib/plot/private/no-gui/plot2d-utils.rkt
@@ -86,21 +86,6 @@
 
   (cond
     [(and x-min x-max y-min y-max)
-
-     (: clip (-> Rect Rect))
-     (define (clip rect)
-       (match-define (vector (ivl rx-min rx-max) (ivl ry-min ry-max)) rect)
-       (define cx-min (if rx-min (max* x-min rx-min) x-min))
-       (define cx-max (if rx-max (min* x-max rx-max) x-max))
-       (define cy-min (if ry-min (max* y-min ry-min) y-min))
-       (define cy-max (if ry-max (min* y-max ry-max) y-max))
-       (let ([cx-min  (min* cx-min cx-max)]
-             [cx-max  (max* cx-min cx-max)]
-             [cy-min  (min* cy-min cy-max)]
-             [cy-max  (max* cy-min cy-max)])
-         (vector (ivl cx-min cx-max)
-                 (ivl cy-min cy-max))))
-
      (flatten-legend-entries
       (for*/list : (Listof (Treeof legend-entry))
         ([rend  (in-list renderer-list)]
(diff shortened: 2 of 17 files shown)


alex-hhh:
With regards to the improvements you mentioned, it is good that you didn't make them here.  Since all tests pass without modification, I know that this pull request does not introduce any bugs in existing code, so I don't need to worry about that.  Once we finish this task, we can add the improvements by themselves and fix the tests.

plot-lib/plot/private/common/contract.rkt
```diff
@@ -34,6 +34,8 @@
 (define anchor/c (one-of/c 'top-left    'top    'top-right
                            'left        'center 'right
                            'bottom-left 'bottom 'bottom-right))
+(define legend-anchor/c (or/c #f anchor/c
+                              (list/c (one-of/c 'inside 'outside) anchor/c)))
```
alex-hhh, line 38: I am not sure I understand your choice for defining the `legend-anchor/c` contract and `Legend-Anchor` type.  I don't understand what the `#f` value is for, and the list definition produces values which don't make sense to me.  E.g. where would the legend be placed when the anchor is `(outside center)` ?  Also, the `inside` entries overlap with the plain one (e.g. `(inside top)` is the same as `top`?)

My thought was to make the legend anchor values a type completely independent from `anchor/c`, some symbols are common, with some additional ones:

```
(define legend-anchor/c (one-of/c 'top-left    'top    'top-right
                           'left        'center 'right
                           'bottom-left 'bottom 'bottom-right
                           'outer-top-left 'outer-center-left 'outer-bottom-left
                           'outer-top-right 'outer-center-right 'outer-bottom-right
))
```

This would allow placing the legend to the left or right of the plot area.  I am not sure that it is worth adding legends at the top and bottom of the plot, did you intend to add that feature?

plot-lib/plot/private/no-gui/plot2d-utils.rkt
```diff
@@ -68,23 +68,29 @@
...
   
   (send area end-renderers)
-  
-  (when (not (empty? legend-entries))
-    (send area draw-legend legend-entries))
-  
   (send area end-plot))
+
+(: get-legend-list (-> (Listof renderer2d) Rect (Listof legend-entry)))
+(define (get-legend-list renderer-list outer-rect)
```
alex-hhh, line 84: I think `get-legend-entries` is a better name for this (and other similar functions in this pull request).  `get-legend-list` seems to indicate that the plot has more than one legend.

alex-hhh (commented):
Overall, the changes look good, my main question is about the choice for the legend anchor type...

plot-lib/plot/private/plot2d/plot-area.rkt
```diff
@@ -672,13 +675,20 @@
...
+      (send pd reset-drawing-params)
+      (when (and (not (empty? legend)) (inside-anchor (plot-legend-anchor)))
+        (draw-legend legend)))
 
     (define/public (draw-legend legend-entries)
       (define gap-size (+ (pen-gap) tick-radius))
+      (define-values (x-min x-max y-min y-max)
+        (if (outside-anchor (plot-legend-anchor))
+            (values dc-x-min (+ dc-x-min dc-x-size)
+                    dc-y-min (+ dc-y-min dc-y-size))
```
alex-hhh, line 687: Do I understand correctly that, if the legend is "outside" you anchor it on the corners of the entire drawingarea?

plot-lib/plot/private/common/contract.rkt
bdeket: - `#f`: This allows to not print the legend at all. Which makes it easy to implement the original question of #49. You now can create a render tree. Pass it to a function that will extract the legend, and later pass this render tree to a plot function (with `#f` as legend anchor) and only plot it without the legend.
- `'(outside top/bottom)`: if the legend is only one item, the legend will be much wider than high. Being able to print it at the top or bottom will reduce the space lost for the actual plot. And it looks better in the top-center than the top-left. So yes this was an intended addition. Especially thinking about a future parameter that allows a horizontal layout for the legend.
- `'(inside anchor/c)` is indeed equivalent with `anchor/c`. When I started I was thinking it might make sense to instead of an anchor, pass coordinates to place it at an exact position. In the end I didn't go further with this.

In general: `'(outside ...)` means it's drawn together with the title, before anything else, and (inside ...) means it's drawn after all other plot items, so indeed `'(outside center)` makes less sense since it will result in a legend in the middle of the width/height of the plot, but below all other drawn items. Whereas `'(inside center)` will result in a legend in the middle of the area-width/area-height of the plot, and above al other items.

Also: according to the documentation anchor/c should allow `'auto`, which it doesn't, and use `'bottom-left` for things that are not a text-label, whereas the legend implementation uses `'top-left`.

plot-lib/plot/private/no-gui/plot2d-utils.rkt
bdeket: I will change it to `get-legend-entry-list`, to make the distinction with the `(Treeof legend-entry)`

plot-lib/plot/private/plot2d/plot-area.rkt
bdeket: This PR doesn't really finish implementing the 'outside functionality. For now it indeed just uses the outside bounds of the entire drawing area and places it relative to that. the last PR (#74) then made the actual space reservation and used better bounds as to not draw over the title. When I made this PR I was anticipating more work in that PR, and since we later split out #75 and #76, There are even less changes in 74...

plot-lib/plot/private/common/contract.rkt
alex-hhh: `'auto` is used for renders such as `point-label`, to select an anchor which ensures that the entire label fits inside the plot area, the value does not make sense for the legend, and I forgot to add the documentation.  There is also a confusion that the `anchor/c` contract is not actually used for the `point-label` and related functions, since they are exported directly from Typed-Racket with the Anchor type -- by the way I use this parameter a lot in my own programs.

the `#f` case is indeed something I did not thing about, but you are correct that we need it if we want to render the legend as a separate pict.  Until now, if the user did not want a legend, they could just not provide a `#:label` entry for any of their renderers.

also, your explanation for `'inside` and `'outside` makes sense -- I didn't think about all the possible cases.  Leave them as they are, and we see if they need changing.

BTW, I think we should add a todo list on #49 for all the small things that we find, to get to them later.  I will add a note about the missing `'auto` documentation as well as the note about not beginning the plot when there are no renderers.

PR #73, opened 2020-09-10
[legend pt4] Plot+legend picts

bdeket:
allow plotting plot and legend to two different picts as was the original question in #49

### DIFF
--- a/plot-lib/plot/private/no-gui/plot2d.rkt
+++ b/plot-lib/plot/private/no-gui/plot2d.rkt
@@ -1,6 +1,6 @@
 #lang typed/racket/base
 
-(require typed/racket/draw typed/racket/class
+(require typed/racket/draw typed/racket/class racket/match
          (only-in typed/pict pict)
          "../common/type-doc.rkt"
          "../common/types.rkt"
@@ -12,6 +12,8 @@
          "../common/nonrenderer.rkt"
          "../common/file-type.rkt"
          "../common/utils.rkt"
+         "../common/plot-device.rkt"
+         "../common/math.rkt"
          "../plot2d/plot-area.rkt"
          "../plot2d/renderer.rkt"
          "plot2d-utils.rkt"
@@ -20,7 +22,7 @@
 
 (unsafe-provide plot/dc
                 plot-bitmap
-                plot-pict
+                plot-pict legend-pict plot+legend-picts
                 plot-file)
 
 ;; ===================================================================================================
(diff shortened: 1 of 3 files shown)


plot-lib/plot/private/no-gui/plot2d.rkt
```diff
@@ -131,6 +133,78 @@
...
+    [(and y-max (not (rational? y-max)))  (fail/kw "#f or rational" '#:y-max y-max)]
+    [else
+     (define saved-values (plot-parameters))
+  
+     (define renderer-list (get-renderer-list renderer-tree))
+     (define bounds-rect (get-bounds-rect renderer-list x-min x-max y-min y-max))
+     (define legend (get-legend-entry-list renderer-list bounds-rect))
+     
+     (define-values (width height)
+       (parameterize/group ([plot-parameters saved-values])
```
alex-hhh, line 158: I don't think we need `parameterize/group` here -- this code runs in the same thread that captured the parameters.

plot-lib/plot/private/no-gui/plot2d.rkt
```diff
@@ -131,6 +133,78 @@
                    #:title title #:x-label x-label #:y-label y-label #:legend-anchor legend-anchor)))
       width height))
 
+(:: legend-pict
+    (->* [(Treeof (U renderer2d nonrenderer))]
+         [#:x-min (U Real #f) #:x-max (U Real #f)
+          #:y-min (U Real #f) #:y-max (U Real #f)]
+         Pict))
+(define (legend-pict renderer-tree
```
alex-hhh, line 141: The name `legend-pict` is a bit confusing (in my opinion anyway).  I think `plot-legend-as-pict` would be a better option, but perhaps there are other possibilities?

I also feel that using the `x-min` ... arguments is a bit counter-intuitive, but I understand why it is necessary.

plot-lib/plot/private/no-gui/plot2d.rkt
```diff
@@ -131,6 +133,78 @@
...
+     (dc (λ (dc x y)
+           (parameterize/group ([plot-parameters  saved-values])
+             (define pd (make-object plot-device% dc x y width height))
+             (send pd reset-drawing-params)
+             (send pd clear)
+             (send pd draw-legend legend (vector (ivl x (+ x width)) (ivl y (+ y height))))
+             (send pd restore-drawing-params)))
+         width height)]))
+
+(:: plot+legend-picts
```
alex-hhh, line 181: I don't think this function is necessary, as it does not provide any more functionality than calling `plot-pict` and `legend-pict` separately, and I would prefer to keep the new API functions to a minimum

plot-lib/plot/private/no-gui/plot3d.rkt
```diff
@@ -171,6 +173,82 @@
                      #:y-label y-label #:z-label z-label #:legend-anchor legend-anchor)))
       width height))
 
+(:: legend3d-pict
+    (->* [(Treeof (U renderer3d nonrenderer))]
+         [#:x-min (U Real #f) #:x-max (U Real #f)
+          #:y-min (U Real #f) #:y-max (U Real #f)]
+         Pict))
+(define (legend3d-pict renderer-tree
```
alex-hhh, line 181: Same remark about the naming, perhaps `plot3d-legend-as-pict`?

plot-lib/plot/private/no-gui/plot3d.rkt
```diff
@@ -171,6 +173,82 @@
...
+    (->* [(Treeof (U renderer3d nonrenderer))]
+         [#:x-min (U Real #f) #:x-max (U Real #f)
+          #:y-min (U Real #f) #:y-max (U Real #f)
+          #:width Positive-Integer
+          #:height Positive-Integer
+          #:title (U String pict #f)
+          #:x-label (U String pict #f)
+          #:y-label (U String pict #f)]
+         (Values Pict Pict)))
+(define (plot+legend3d-picts renderer-tree
```
alex-hhh, line 234: Same remark as for 2D plots -- I don't think this function is necessary

alex-hhh (commented):
I have added some practical commends about the pull request, and I will merge it once they are addressed, however, I am a concerned about the entire idea of providing the plot legend as a picture.

I know this was requested in the original issue, but the forum discussions requested just placing the legend outside the plot area, and the idea of having the legend as a pict was the idea of a single person and I feel it was not discussed enough.

I am not sure that anyone would use this feature, so it is very difficult to determine if we provided the right thing.  And we have limited opportunity to correct things once they are published in a Racket release, as the plot package needs to be backwards compatible forever.

However, placing the legend outside the plot area is a feature that is available in other plot libraries and we can be more certain that we provided the right (or at least a workable) interface...

bdeket:
I don't mind not introducing this, as I prefer to have the legend in the plot. But if we do, I think the better option is the one where plot+legend are provided, so at least the `#:x-min` etc arguments make more sense. Feel free to close this PR.

alex-hhh:
Unless someone has a concrete use case for this, as opposed to just brainstorming some possible use cases, we can easily add it in, given all the refactoring that was done.   However, once we add something to the plot package API, it will need to be maintained practically forever, so I am hesitant to add things for "just in case" scenarios, because we might implement them to support only limited use cases so they have the potential to be useless to most of the users.

Also note, that users can also create manual legends using picts, although perhaps a bit inconvenient, but with a lot of flexibilities, this was illustrated in the google-groups discussion listed from #49.

Metaxal:
I'd like to try this change, but the branch has been deleted and the `git apply patc 73.patch` fails with patch does not apply.

I could copy/paste the code manually, but I'd rather avoid as I expect complications. Can the branch be restored, or is there another way I can patch this?

bdeket:
I deleted this branch on all my devices. It seems github gives me an option to "restore branch", but I don't know if it will work.
One thing to keep in mind with these patches for the legend: there where a bunch of them at the same time that were interconnected. And after dropping this branch updates for the other branches where not incorporated into this one.
I guess the best thing to first try is reverting to commit e97c8e74dcd7ba68e971ecfcbde5dd14224e14ee or somewhere thereabout and see if at this point the patch still works.

Metaxal:
Quickly: I could eventually patch it on top of HEAD (sorry, just saw you restored the branch), and it works fine with `ht-append` but weirdly offsets the legend by the plot amount with `hc-append`, and double that with `hb-append`. This can be fixed by using `freeze` on the legend first, but it shouldn't be the case I think.

I'll get back to this soon, hopefully.

Thanks for starting this!

PR #75, opened 2020-09-14
Fixpoint margin

bdeket:
Currently the margin calculation as used in 2d/plot-area counts the initial margin double. Leading to early opt out for finding margins. This PR addresses this.

---

```
(require plot pict)
(plot (function values 0 1)
      #:title (text "a realy realy ridiculously high title" null 12 1.55))
```
untitled


alex-hhh:
For future reference, to use text picts in tests, you need to setup a "mock drawing context", here is an example: https://github.com/racket/plot/blob/e97c8e74dcd7ba68e971ecfcbde5dd14224e14ee/plot-test/plot/tests/PRs/58.rkt#L16

PR #78, opened 2020-09-16
Legend: tests

bdeket:
Tests for changes for #49 

In my opinion there is missing a test for the interactive handling of plot-snip%.
I imagine it to be something like (3d):
1. draw to a dc
2. set `plot-animating?` parameter to `#t`
3. change `plot3d-angle` and/or `plot3d-altitude` parameters of plot-snip% and draw again
4. set `plot-animating?` back to `#f` and draw again


alex-hhh:
I pushed a different set of tests which cover all plot legend placement positions for both 2d and 3d plots.  You are correct that we are missing some test coverage with respect to plot snips (and indeed I fixed a bug related to that in commit cbc0620ce851baa93add786d65018ecbacbf82e5, but for now, it is simpler to add the snip tests on a case-by-case basis (i.e. when actual bugs are found).

bdeket:
I mentioned this here because this was one bug I introduced that I only discovered by chance. And then it took me some time to track down what went wrong.
I was looking into the test package to see how I could test this, but I didn't come up with a solution.

alex-hhh:
https://github.com/racket/plot/blob/master/plot-test/plot/tests/PRs/snip-empty-selection.rkt and https://github.com/racket/plot/blob/master/plot-test/plot/tests/PRs/snip-overlay-renderers.rkt show how to test some plot snips (they test specific bugs).  

The idea is that you can create a plot snip and you'll have to register it, but than you can send it mouse commands, for example to rotate and draw commands to obtain the plot image.  However, depending on what you are trying to do, this might get complex.  If you want to have a go, you can of course try and I will try to help you out as best as I can.

PR #85, opened 2021-03-01
minor fix for intro doc

xxyzz:
I also wanna ask why `plot3d`'s output feels weird when `#:altitude` is not `25`:

```racket
(plot3d (polar3d (λ (θ ρ) 1) #:color 2 #:line-style 'transparent)
        #:altitude 0 #:angle 0)
```

Shouldn't this code output a circle instead of an ellipse?

### DIFF
--- a/plot-doc/plot/scribblings/intro.scrbl
+++ b/plot-doc/plot/scribblings/intro.scrbl
@@ -105,7 +105,7 @@ Clearly, the blue-colored interval between sine waves is drawn first.
 
 @section{Renderer and Plot Bounds}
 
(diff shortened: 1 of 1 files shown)


alex-hhh:
The plot package will include an `#:aspect-ratio` parameter which will help with controlling the aspect ratio of the plot area (although  it does not work 100% correctly for 3D plots).

As for the actual documentation fix, I will close this pull request, as the change is too small and I want to discourage people from submitting low effort changes -- I have limited time as a maintainer and even such a small change takes a large chunk of my already tight time.

If  you want to help out you can either:

* proofread and spell-check the plot manual (or a meaningful chunk of it)
* write the documentation for the new plot metrics feature (see #90 and related issues)

PR #86, opened 2021-03-03
add scientific? argument to log-ticks and log-ticks-format

ralsei:
Similarly to `linear-ticks` and `linear-ticks-format`, this adds a keyword argument `scientific?` to these functions to disable scientific numbers in tick labels. Not much to say about this, really.


alex-hhh:
Can you add some tests for these changes, there are several examples in https://github.com/racket/plot/tree/master/plot-test/plot/tests/PRs.  You can just name your test after the pull request number (i.e. 86.rkt).   The test should have sample plots both with `#:scientific?` set to `#t` and `#f`.   Let me know if you need help with the tests.  Thanks.

ralsei:
Tests written. (The rationale for the `function` is because a logarithmic graph should be linear after a log transform).

alex-hhh:
Thanks for doing the changes.  Unfortunately, I have just noticed that the build is in fact broken, as it somehow manages to build and test the plot package shipped with the distribution, instead of the changes being tested.  As such, your changes are not in fact built and your tests not run.  I will need to investigate why this is the case and I'll hold back merging this until that problem is fixed.

alex-hhh:
Now that #87 was merged, can you please pull from the master branch into this one, so we get a build using your changes?  Thanks , Alex.

alex-hhh:
Thanks for working on this.

PR #87, opened 2021-03-04
make CI build from the repository and not from the catalog

ralsei:
This likely also fixes the CI issue from #86 (...somehow). Looking through the logs of the CI build:
```
Resolved "plot" via file:///home/runner/work/plot/plot/catalog
00: Resolved "plot-lib" via file:///home/runner/work/plot/plot/catalog
00: Resolved "plot-gui-lib" via file:///home/runner/work/plot/plot/catalog
Resolved "plot-doc" via file:///home/runner/work/plot/plot/catalog
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,11 +9,11 @@ jobs:
       # the failure is specific to a Racket version.
       fail-fast: false
       matrix:
-        racket-version: ["7.9", "current"]
-        racket-variant: ["regular", "CS"]
+        racket-version: ["8.0", "current"]
+        racket-variant: ["BC", "CS"]
     steps:
       - uses: actions/checkout@v2
-      - uses: Bogdanp/setup-racket@v0.10
+      - uses: Bogdanp/setup-racket@v1.1
         with:
           architecture: x64
           distribution: minimal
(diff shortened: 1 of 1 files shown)


ralsei:
And, for whatever reason, it's now pulling from the catalog. Wonderful...

alex-hhh:
Looking at the build logs for the "Racket 8.0 CS", I see:

```
Resolving "plot" via https://download.racket-lang.org/releases/8.0/catalog/
Downloading https://download.racket-lang.org/releases/8.0/pkgs/plot.zip
00: Resolving "plot-lib" via https://download.racket-lang.org/releases/8.0/catalog/
01: Resolving "plot-gui-lib" via https://download.racket-lang.org/releases/8.0/catalog/
Resolving "plot-doc" via https://download.racket-lang.org/releases/8.0/catalog/
```

So this change still resolves plot via the catalog, instead of using the local checked out sources.

My guess is that the problem is caused by the fact that the catalog setup commands are run as a regular user (no sudo), while the build commands are run using "sudo", so they might pick up a different config location, for the root user.  I think this worked in an older version of setup-racket, and got "fixed" in newer versions, but I didn't notice  it.

I think the fix is to remove the "sudo" part from the build commands -- the build should work without them.  Can you please try that?  

Updating to 8.0 and to racket-setup 1.1. is also good, so please keep these changes in the pull request.

Thanks,
Alex.

ralsei:
It looks like making it rootless was the trick.

alex-hhh:
Thanks for trying this out and submitting a fix.

Issue #84, opened 2021-02-17
Inverting axis transforms


dented42:
For completeness it seems like the following functions for reversing axis transforms should exist:
```
;;; Turn an axis transform inside out
(: axis-transform-invert (Axis-Transform -> Axis-Transform))

;;; apply an axis transform in reverse
(: apply-axis-transform/reverse (Axis-Transform Real Real -> Invertible-Function))
```

alex-hhh:
I am not sure how any of these functions would be implemented, but if you provide an implementation as a pull request, I am happy to review it and help you with getting it ready for merging (documentation , tests, etc).

Also, would you be able to provide some example use cases for these functions?

alex-hhh:
Hi @dented42 , with reference to my previous message,  can you please:

* provide some examples where the functions would be useful
* provide an explanation on how these functions would be implemented?  A sample Racket implementation would be great.

If you consider providing a pull request, I can help you with getting it ready for merging (documentation , tests, etc).  This is not necessary, but I cannot see how these functions can be implemented, so I cannot do anything about this issue.

dented42:
What made me think of it initially was mostly that axis transforms (If I'm understanding the implementation correctly) seem to be made of invertible functions, and it seems aesthetically odd that the invertibility isn't exposed for completeness.

The specific example I had in mind was wanting an exponential transform, and inverting the log transform seemed like the most straightforward way.

I'll try to whip up an implementation when I'm able.

dented42:
(Just noticed I mucked up one of the signatures in my initial post, I've corrected it)

dented42:
Another more pressing example that has come up is actually trying to integrate `plot` into a GUI (I've been playing with `lux` lately), and wanting to translate mouse/window coordinates back into plot coordinates. Besides using `lux` it also just seems useful to be able to draw a plot into a graphics context and be able to map back and forth between canvas coordinates and the coordinate system defined in the plot. Obviously the axis transform is only a single part of that equation, but it seems like an important part as it's the part that introduces nonlinearity.

alex-hhh:
The functionality you requested in #83 is already available for plot snips, where the coordinates passed back to the callback are converted back to plot coordinates, and this works in the presence of axis transforms.  So these functions are not needed to implement such functionality.

Also, #83 is about how to expose already existing functionality to the other plot entry points (such as `plot/dc` and related functions), which are designed for non-interactive use.  As you see from that discussion, the design of that API is not easy...

[...]

PR #88, opened 2021-03-08
Add support for specifying the aspect ratio of the plot area

alex-hhh:
Plot will calculate the plot area such that it fills up the entire available size of the plot, but this creates a problem for plots which need to maintain their "shape".  For example, by default when plotting a circle, you get an ellipse:

```racket
(require plot)
(plot-background "lightblue")
(plot-x-label #f)
(plot-y-label #f)

(plot (polar (lambda (t) 1)) #:width 400 #:height 200)
```

circle-non-ar

This pull request introduces the `#:aspect-ratio` parameter for all plot functions and a corresponding `plot-aspect-ratio` parameter, which allows specifying the ratio between the width and height of the plot  area.  For example, by specifying an aspect ratio of 1, the plot will actually look like a circle.

```racket
(plot (polar (lambda (t) 1)) #:width 400 #:height 200 #:aspect-ratio 1/1)
```

circle-with-ar

The aspect ratio setting also works for 3D plots.


bdeket:
Hi,
I didn't manage to test it yet and will do so tonight, but if I understand correctly the aspect ratio is defined in terms of the plot-area width and height.
I would like to suggest to instead define this in terms of the x and y (and z) axis so that when you are drawing an ellipse you don't need to calculate the right aspect ratio every time you change the shape.
An alternative is to add an option axis-constrained? ( I think maple did this ) so you have only two options: maximize the available area or keep axiss with same unit step size

alex-hhh:
Hi @bdeket , thanks for the feedback.  Yes the aspect ratio is defined in terms of the plot area, this seemed more natural to me, especially when using aspect ratios of 4:3 and 16:10.  It is also simpler to think about this for 3d plots, where the aspect ratio applies to the view.

Just to make sure I understand what you are suggesting, since I am not familiar with maple.  Are you suggesting that the aspect ratio would be defined in terms of plot units, such that, for example, with an aspect ratio of 1, this plot would remain a circle, even though the x-axis range is 4 but the y axis range is 2?

```
(plot (polar (lambda (t) 1)) #:x-min -2 #:x-max 2 #:y-min -1 #:y-max 1)
```

bdeket:
Indeed, that was the idea.
I suggested this because the only use case I could think of to use the aspect-ratio is exactly that: keeping plot units in x and y (z) direction exactly in step.
But maybe you have another use case, and this question is orthogonal to that.

I think in maple this same ratio for the plot units was also enforced when you started zooming in and out, but I'm not sure anymore (I would consider it a plus)

samth:
A different approach here might be to allow controlling two kinds of widths/heights separately -- the plot itself and the entire surface. That would allow achieving your result by specifying the plotted area to have size 200x200. This incidentally might be useful for creating plots that are aligned when combined as picts.

rfindler:
Just to chime in with @samth 's comments, one of the things I have struggled with when using plot (mostly everythign with plot is amazingly smooth sailing!) is when I have multiple plots that I want to put into the same figure and I want the mapping of plot coordinates -> drawn coordinates to be the same for all the plots and I want to put a bunch of plots together into a figure (by doing stuff like `hbl-append` with the picts that come out of the plot library). Things like the axis labels being different sizes foils me, however. If these changes help with that usecase, that'd be fantastic. Also, a related request might be a few examples that explain how to make this usecase work well in the docs would also be very welcome so people who are using plot for papers find it and know what to do.

bdeket:
wrt "maintaining shape"
I think I still don't completely understand the intent. Is this to address the comment of #85 or to address #7 , or something else entirely?

```
(plot3d (polar3d (λ (ϕ θ) 1)) #:width 400 #:height 200 #:altitude 0)
```
b1

```
(plot3d (polar3d (λ (ϕ θ) 1)) #:width 400 #:height 200  #:altitude 0 #:aspect-ratio 1/1)
```
b2

If the aspect ratio was in terms of plot-units, the second picture would show a ball(/circle) whatever the view angle

---

Also, when I start zooming in on your second circle (aspect ratio 1/1) it looks like something goes wrong with calculating the optimum size (too much white space at bottom and left):
zoomed

alex-hhh:
I understand that many of people want #7 and #84 to be implemented and that many people would like to be able to align the plot areas of several plots, for the purpose of writing publications.   This pull request does not address any of those issues.   Those issues are constantly on my mind, but as of today, I don't know how to implement them in a way that it does not create more problems than it solves.  The main problem with them is that the plot package has many use cases, and so far, all proposed solutions handle one or two use cases well, while creating potential problems for other use cases.

----

The purpose of this feature is to address the question in #85, as well as the problem I had when I wrote my [Plot Animations](https://alex-hhh.github.io/2021/01/plot-animations.html) blog post, where I had to manually adjust the size of the plot, until the circle looked like a circle.  As @bdeket noted, #85 is not addressed perfectly...  

To a certain extent can also be used as an example of how many files have to change to add one single parameter to the plot API, which I think is one drawback of the current plot API...

The problem is addressed in such a way that, I feel it is easy for someone with a programming background to use the feature and it works reasonably well with all use cases of the plot package.  In particular:

* I would expect an aspect ratio of 1:1 to result in a square plot area, and I think this is what most people would expect.  If the aspect ratio would be in plot units, as @bdeket suggested, a plot with an X range of 2 and an Y range of 1 would be a rectangle with an aspect ratio of 1:1, which is unintuitive.  @bdeket your suggestion of maintaining "square" plot units is also useful, and perhaps could be added as a separate plot parameter, perhaps in a different PR
* specifying the plot area size separately, as @samth suggested, will put additional burden on the user to control the layout, and there will be no help from the plot package in determining what the optimal fill of the entire plot image would be, especially when taking into account labels and axes.  The size of the plot area would have to be constantly tweaked as the plot data (and therefore labels) change -- this might be acceptable when creating plots for a publication, where the data is fixed, but won't work in any other use case.  It is also unclear how the plot area would grow when a plot frame created by `plot-frame` is resized by the user.

@bdeket , I will look into the issue you noticed with the "deep zooming"

rfindler:
Indeed. My bad @alex-hhh . Sorry.

alex-hhh:
Hi @bdeket , this is the result of zooming in excessively with the current plot package (without the changes in this PR).  It looks like something is wrong with calculating the plot area size, but it is not introduced by this PR.

```racket
(require plot)
(plot-background "lightblue")
(plot-x-label #f)
(plot-y-label #f)

(plot (polar (lambda (t) 1)) #:width 400 #:height 200)
```

image

alex-hhh:
With regards to the 3D plots being distorted, this is caused by the fact that the projection itself is not considering the aspect ratio, I had a brief look at the code in plot-area.rkt for the 3D plots and it looks like the code is doing "manual" calculations instead of defining "world->view", "view->projection" and "projection->screen" matrices, so it is not easy to change to take the aspect ratio into account -- it would be a nice improvement to rewrite this aspect of the 3D plots.  This would also make 3D plots faster, since going from 3D world coordinates to DC coordinates would be a single matrix multiplication instead of a chain of transformations, as it is now.

This refactoring would also have to be done if we implement the "plot units" based aspect ratio.

PR #89, opened 2021-03-09
correct calculation text extents

bdeket:
correct calculation text extents for (member anchor '(left center right))

minimal example:
```
#lang racket/base

(require plot)
(plot-background "lightblue")
(plot-x-label #f)
(plot-y-label #f)

(plot (polar (lambda (t) 1))
      #:x-min 4025584149602844706/10000000000000000000
      #:x-max 4025584149602844712/10000000000000000000
      #:y-min 6805367900508167/10000000000000000
      #:y-max 6805367900508171/10000000000000000
      #:width 400 #:height 200)
```
ext1

with PR
ext2


alex-hhh:
Thanks for looking into this bug and submitting a fix.

Issue #83, opened 2021-02-17
Expose converting between plot/view/context coordinates


dented42:
I'm trying to integrate plot into a larger program and need a way to take coordinates in the drawing context (I'm using `plot/dc`) and convert them to coordinates in the plot (and back again).

I'm aware of the word done with `set-mouse-event-callback` etc. but I don't think using snips makes sense for my program (I'm using `lux`, which just gives me a drawing context).

dented42:
Given how `plot` is architected I'm unsure of the best way to do this. I have made a list of (possibly terrible) ideas.

- have a `plot` variant that instead of drawing a plot returns a value/object that can be interrogated about various metrics of the plot that would be drawn?
- have `plot/dc` optionally return a value/object that exposes various metrics
- have `plot/dc` optionally take a callback that is passed a metrics value and does something with it? (this seems like it might be the easiest to implement but the most annoyingly imperative)

A metrics objects might be a struct or object along the lines of:
```
(struct plot-metrics (x-min         ; real
                      x-max         ; real
                      y-min         ; real
                      y-max.        ; real
                      plot->view/x  ; invertible-function
                      view->dc/x    ; invertible-function
                      plot->view/y  ; invertible-function
                      view->dc/y )) ; invertible-function
```

alex-hhh:
The main challenge with this request (and #7 as well) is designing the new API -- unfortunately, the design of the plot package does not make this easy, since there are many entry points to creating plots, and all of them need to support all the features (if it makes sense for them, of course).

Returning a metrics object would work for `plot/dc` since it currently returns `void`, so the API change will likely be backwards compatible, but a similar request was made here for plots which produce `picts`: https://github.com/racket/plot/issues/7#issuecomment-650174050.  In particular, this feature would need to work with:

* `plot/dc`
* `plot-pict`
* `plot-bitmap`
* `plot-snip`

In addition to that, there are some special `plot` functions in `plot/pict` and `plot/bitmap` intended for use during Scribble and web page rendering, and perhaps these would need extending too...

I think the main `plot` , along with `plot-frame` and `plot-file` would not need to support this functionality, but I may be wrong.

And, of course, it would also be nice if the API change would be made in such a way that the 3D variants of these functions could also be updated to provide similar functionality.

rfindler:
[...]

PR #90, opened 2021-03-18
Plotmetrics

bdeket:
Possible implementation for getting plot-metrics from plots. see #83 

#### Currently implemented methods:
- `get-plot-boundaries`: `(-> (Vectorof (Vector Real Real))`  - give min/max x y (z) values
- `plot->dc`: `(-> (Vecorof Real) (Vectorof Real))` - translate plot x y (z) coordinates to dc x y coordinates

#### Possible problems:
- This changes the return type for `plot/dc`, `plot-bitmap`, `plot-snip`, `plot-pict`
- typed-untyped interaction: Except for `plot-pict` the return types are new objects for which no interface or class is defined. There will be no straighforward way to check if the returned objects implement `Plot-Metrics<%>`

#### todo:
- [ ] documentation
- [ ] tests

### DIFF
--- a/plot-lib/plot/private/common/types.rkt
+++ b/plot-lib/plot/private/common/types.rkt
@@ -162,3 +162,17 @@
    [draw-pict (->* [pict (Vectorof Real)] (Anchor Real) Void)]
    [calculate-legend-rect (-> (Listof legend-entry) Rect Anchor Rect)]
    [draw-legend (-> (Listof legend-entry) Rect Void)]))
+
+(define-type Plot-Metrics<%>
+  (Class
+   [get-plot-bounds (-> (Vectorof (Vector Real Real)))]
+   [plot->dc (-> (Vectorof Real) (Vectorof Real))]))
+
+(struct plotpict pict ([bounds : (Vectorof (Vector Real Real))]
+                        [plot->dc : (-> (Vector Real Real) (Vectorof Real))]))
+(define (pict->pp [P : pict]
+                  [bounds : (Vectorof (Vector Real Real))]
+                  [->dc : (-> (Vector Real Real) (Vectorof Real))]) : plotpict
+  (plotpict (pict-draw P) (pict-width P) (pict-height P) (pict-ascent P) (pict-descent P) (pict-children P) (pict-panbox P) (pict-last P)
+             bounds ->dc))
+
--- a/plot-lib/plot/private/common/types.rkt
+++ b/plot-lib/plot/private/common/types.rkt
@@ -170,6 +170,7 @@
 
 (struct plotpict pict ([bounds : (Vectorof (Vector Real Real))]
                         [plot->dc : (-> (Vector Real Real) (Vectorof Real))]))
+(define-type PlotPict plotpict)
 (define (pict->pp [P : pict]
                   [bounds : (Vectorof (Vector Real Real))]
                   [->dc : (-> (Vector Real Real) (Vectorof Real))]) : plotpict
--- a/plot-lib/plot/private/common/types.rkt
+++ b/plot-lib/plot/private/common/types.rkt
@@ -163,17 +163,5 @@
    [calculate-legend-rect (-> (Listof legend-entry) Rect Anchor Rect)]
    [draw-legend (-> (Listof legend-entry) Rect Void)]))
 
-(define-type Plot-Metrics<%>
-  (Class
-   [get-plot-bounds (-> (Vectorof (Vector Real Real)))]
-   [plot->dc (-> (Vectorof Real) (Vectorof Real))]))
-
-(struct plotpict pict ([bounds : (Vectorof (Vector Real Real))]
-                        [plot->dc : (-> (Vector Real Real) (Vectorof Real))]))
-(define-type PlotPict plotpict)
-(define (pict->pp [P : pict]
-                  [bounds : (Vectorof (Vector Real Real))]
-                  [->dc : (-> (Vector Real Real) (Vectorof Real))]) : plotpict
-  (plotpict (pict-draw P) (pict-width P) (pict-height P) (pict-ascent P) (pict-descent P) (pict-children P) (pict-panbox P) (pict-last P)
-             bounds ->dc))
-
+(require "plotmetrics.rkt")
+(provide Plot-Metrics<%> Plot-Pict Plot-Metrics-Functions)
--- a/plot-lib/plot/private/no-gui/plot2d.rkt
+++ b/plot-lib/plot/private/no-gui/plot2d.rkt
@@ -1,7 +1,7 @@
 #lang typed/racket/base
 
 (require typed/racket/draw typed/racket/class
-         (only-in typed/pict pict)
+         typed/pict
          "../common/type-doc.rkt"
          "../common/types.rkt"
          "../common/draw.rkt"
@@ -15,16 +15,37 @@
          "../plot2d/plot-area.rkt"
          "../plot2d/renderer.rkt"
          "plot2d-utils.rkt"
-         "evil.rkt"
+         "../common/math.rkt"
+         (only-in "evil.rkt" new-post-script-dc% new-pdf-dc% new-svg-dc%)
          typed/racket/unsafe)
 
 (unsafe-provide plot/dc
                 plot-bitmap
                 plot-pict
-                plot-file)
+                plot-file
+                plotpict-bounds
+                plotpict-plot->dc)
+
+;; ===================================================================================================
+;; Extra types
+
+(define-type Plot-Metrics<%>
+  (Class
+   [get-plot-bounds (-> (Vector (Vector Real Real) (Vector Real Real)))]
+   [plot->dc (-> (Vectorof Real) (Vectorof Real))]))
+
+(struct plotpict pict ([bounds : (Vector (Vector Real Real) (Vector Real Real))]
+                        [plot->dc : (-> (Vector Real Real) (Vectorof Real))]))
+(define (pict->pp [P : pict]
+                  [bounds : (Vector (Vector Real Real) (Vector Real Real))]
+                  [->dc : (-> (Vector Real Real) (Vectorof Real))]) : plotpict
+  (plotpict (pict-draw P) (pict-width P) (pict-height P) (pict-ascent P) (pict-descent P) (pict-children P) (pict-panbox P) (pict-last P)
+             bounds ->dc))
+
 
 ;; ===================================================================================================
 ;; Plot to a given device context
+
 
 (:: plot/dc
     (->* [(Treeof (U renderer2d nonrenderer))
@@ -37,7 +58,7 @@
           #:y-label (U String pict #f)
           #:aspect-ratio (U Nonnegative-Real #f)
           #:legend-anchor Legend-Anchor]
-         Void))
+         (Instance Plot-Metrics<%>)))
 (define (plot/dc renderer-tree dc x y width height
                  #:x-min [x-min #f] #:x-max [x-max #f]
                  #:y-min [y-min #f] #:y-max [y-max #f]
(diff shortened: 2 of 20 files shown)


alex-hhh:
Thanks for starting the work on this feature.  Unfortunately, I will only have time to look at this in detail next week, but it would be good to get the rest of the Racket teams opinion on this approach.  My own take is that the overall approach is OK, and it does solve some of the challenges with returning extra information with the plot itself.

samth:
A couple thoughts.

1. It seems like it would be easy to add an interface with the two relevant methods to all the new classes and make it easy to test for in untyped code that way.
2. I think an abstraction for creating the classes might be nice, something like
```
(define (plot-metrics-mixin c bounds ->dc)
    (class c (plot-metrics<%>) 
        (super-new)
        (define/public (get-plot-bounds) (bounds)
        (define/public (plot->dc [v : (Vectorof Real)]) (->dc v))))
```

bdeket:
@samth Currently `plot/dc` and `plot-bitmap` are defined in TR, and `plot-snip` is defined in untyped racket. That makes it hard to do what you are suggesting, right? (This was also the reason of my question about interfaces and TR yesterday)

alex-hhh:
Hi @samth, is it OK to change the return type of `plot/dc` from `void` to a `plot-metrics<%>` object?  I think it is OK, but @rfindler indicated in ths comment that it might not be a good idea, but did not explain why: https://github.com/racket/plot/issues/83#issuecomment-781037659

Another questions for @samth: this is a complex feature, and it is not clear to me that these changes will be sufficient or even useful for the types of things people have requested them.  I don't think we can get this right in one go, since we have only one use cases:

* ideally, I would like @dented42 to try these changes out, but I am not sure if he/she still has time and he/she is interested in this
* finish and merge this implementation to the main branch and add it to a Racket release, so other people can experiment with it
* mark this feature as experimental in the Racket release, and allow changing these APIs in non-backwards compatible ways based on feedback from users who try to use these features 

What do you think?

plot-lib/plot/private/common/types.rkt
```diff
@@ -162,3 +162,18 @@
    [draw-pict (->* [pict (Vectorof Real)] (Anchor Real) Void)]
    [calculate-legend-rect (-> (Listof legend-entry) Rect Anchor Rect)]
    [draw-legend (-> (Listof legend-entry) Rect Void)]))
+
+(define-type Plot-Metrics<%>
+  (Class
+   [get-plot-bounds (-> (Vectorof (Vector Real Real)))]
+   [plot->dc (-> (Vectorof Real) (Vectorof Real))]))
+
+(struct plotpict pict ([bounds : (Vectorof (Vector Real Real))]
```
alex-hhh, line 171: I think the structure should be named `plot-pict` as this is more consistent with other names.

plot-lib/plot/private/common/types.rkt
```diff
@@ -162,3 +162,18 @@
    [draw-pict (->* [pict (Vectorof Real)] (Anchor Real) Void)]
    [calculate-legend-rect (-> (Listof legend-entry) Rect Anchor Rect)]
    [draw-legend (-> (Listof legend-entry) Rect Void)]))
+
+(define-type Plot-Metrics<%>
+  (Class
+   [get-plot-bounds (-> (Vectorof (Vector Real Real)))]
+   [plot->dc (-> (Vectorof Real) (Vectorof Real))]))
```
alex-hhh, line 169: We also need a `dc->plot` method, since that would be used to convert mouse coordinates to the plot space.

plot-lib/plot/private/no-gui/plot2d.rkt
```diff
@@ -68,7 +70,16 @@
                     [plot-legend-anchor  legend-anchor])
        (define area (make-object 2d-plot-area%
                       bounds-rect x-ticks x-far-ticks y-ticks y-far-ticks legend dc x y width height))
-       (plot-area area renderer-list))]))
+       (plot-area area renderer-list)
+
+       (define bounds (vector (vector (assert (ivl-min (vector-ref bounds-rect 0)) real?)
```
alex-hhh, line 75: I think the bounds should be calculated and memory allocated for them in the `get-plot-bounds` method below, perhaps lazily initialize them as a member of the returned plot-metrics object.  

The `plot/dc` method is used in every 2D plot and most of these uses will just discard the plot-metrics object  so I want to avoid the memory allocation penalty (and runtime execution penalty) in all those use cases.

plot-lib/plot/private/no-gui/plot2d.rkt
```diff
@@ -68,7 +70,16 @@
...
+                                      (assert (ivl-max (vector-ref bounds-rect 0)) real?))
+                              (vector (assert (ivl-min (vector-ref bounds-rect 1)) real?)
+                                      (assert (ivl-max (vector-ref bounds-rect 1)) real?))))
+       (new (class object%
```
alex-hhh, line 79: While creating a new class type and immediately instantiating an object is supported by Racket, I am not sure if this is the most efficient way to do it in terms of memory allocation and runtime performance -- this object will be discarded in most current uses of the plot package).

@samth do you know whether it would be more efficient to create a separate `plot-metrics%` class and just instantiate this here?

plot-lib/plot/private/no-gui/plot3d.rkt
```diff
@@ -86,7 +87,18 @@
                       bounds-rect x-ticks x-far-ticks y-ticks y-far-ticks z-ticks z-far-ticks
                       legend-list
                       dc x y width height))
-       (plot-area area renderer-list))]))
+       (plot-area area renderer-list)
+
+       (define bounds (vector (vector (assert (ivl-min (vector-ref bounds-rect 0)) real?)
```
alex-hhh, line 92: same notes about allocating `bounds` as for the 2D plot.

plot-lib/plot/private/no-gui/plot3d.rkt
```diff
@@ -86,7 +87,18 @@
...
+                                      (assert (ivl-max (vector-ref bounds-rect 0)) real?))
+                              (vector (assert (ivl-min (vector-ref bounds-rect 1)) real?)
+                                      (assert (ivl-max (vector-ref bounds-rect 1)) real?))
+                              (vector (assert (ivl-min (vector-ref bounds-rect 2)) real?)
+                                      (assert (ivl-max (vector-ref bounds-rect 2)) real?))))
+       (new (class object%
```
alex-hhh, line 98: ... and for creating a new class and instantiating it...

plot-lib/plot/private/no-gui/plot2d.rkt
samth: @mflatt will know better here, but it will at most save a bit of allocation here.

samth:
@bdeket I don't really understand your comment. You can define the interface in untyped code, and then ensure that it's implemented by all the classes. There would be some duplication, which is unavoidable here for the reasons I described on Slack, but that doesn't seem so bad here.

samth:
@alex-hhh On the backwards-compatibility issue, I think it's unlikely to be a problem. I agree with @rfindler (where did his comment go?) that the most likely problem is someone with a contract that starts failing, but I still think that's unlikely.

On the experimental issue, I'm less sure. We should think about what changes we might want to make that wouldn't be possible in a compatible way (say, by adding another method).

bdeket:
@samth I didn't understand how to do what you were proposing. But I think I do now:
Create interface and a mixin (or otherwise class) in untyped racket.
`require-typed` them in the TR modules and apply them there.

My initial thought was only having the interface in untyped code, and I didn't see how to integrate that with the other pieces.

plot-lib/plot/private/common/types.rkt
bdeket: What would be the return type be for plot3D, a line?

bdeket: Or alternatively: apart from adding `dc->plot`, also provide a `plane-vector` method. For 2D this would always return #(0 0 1), for 3D this would return the unit vector `e` that is perpendicular to the screen. So for `dc->plot` 3D can just return the coordinate `c0` for the plane trough the origin. Then the user can reconstruct the right coordinate by adding these two together: `(+ c0 (* a e))`

plot-lib/plot/private/no-gui/plot2d.rkt
alex-hhh: We only have one use case for these changes (https://github.com/racket/plot/issues/83#issuecomment-790017560) and this would involve calling `plot/dc` several times a second (30FPS perhaps).  I think this code should avoid any unnecessary memory allocations and defer any deferrable computations until they are needed (e.g. calculating the bounds).

plot-lib/plot/private/common/types.rkt
alex-hhh: oops, `plot-pict` is a function name...

plot-lib/plot/private/common/types.rkt
alex-hhh: I did not think about the 3D case and the 2D plot area already has a `dc->plot` method.   

I am thinking about this feature in terms of "what functionality is needed to construct interactive plots?" since this was the original use case for creating #83.  I have some experience with this, from using the `plot-snip%` interface for 2D plots, and I know that there needs some way from converting from DC to plot coordinates, so be able to interact with the mouse.

For 3D plots, I have no experience with them, and not sure what is needed to achieve the final goal.  Making `dc->plot` return a line is the mathematically correct thing to do, but I am not sure if this would be sufficient and/or useful for anything practical.

Perhaps we could not do the 3D plot changes at all, and wait until someone comes with a use case that we can use to validate the design?

alex-hhh:
Hi @samth 

> On the experimental issue, I'm less sure. We should think about what changes we might want to make that wouldn't be possible in a compatible way (say, by adding another method).

My concern is that this is a large and complex change, and we only have only one use case for it (https://github.com/racket/plot/issues/83#issuecomment-790017560).  There is a distinct possibility that, apart from perhaps a few tests and toy examples, no one will use this, and, when they will, we'll discover that, despite our best efforts, the interface is not sufficient or not useful.

As such, I would like to get these changes out in a Racket release (since most people won't build the plot package themselves), announce them, allow people to try them out and provide feedback then incorporate that feedback in a future plot package release.

alex-hhh:
Hi @bdeket , I think the changes look good.  Can you please add some unit tests for these, I think we should leave the documentation update for a separate pull request, as this one is getting too large.

bdeket:
Hi @alex-hhh, I will have time tomorrow/later this week.
But I was wondering, regarding trying out this changes in a release. Would it make sense to setup a `plot/unstable` where this change could be tested for some time, before being merged into the main branch. The bad part is, that it will involve some code duplication

alex-hhh:
I also thought about creating a `plot/unstable` module for these changes, but it might not be worth the effort, since it requires a lot of changes.

Another option is to not merge this branch yet, and, when someone asks for this feature in the future, point them to this branch and complete the work at that time based on their feedback.  The plot package does not change that often and it would be simple to keep this branch up to date with the main branch.  I am not sure how many people would use this feature in the short to medium term (say 6 - 12 months), the person who originally created the issue has not responded yet, and it may be that they are no longer interested.

Finally, I am also happy to just merge this.  When someone starts using these features, we'll get some feedback and decide at that time what to do if the API needs to be changed.

mfelleisen:
Don't create an `unstable` module/directory.  We had one at the top level and it leads to the usual problems.

plot-test/plot/tests/PRs/90.rkt
```diff
@@ -0,0 +1,81 @@
...
+                        (check-within (send ps plot->dc (send ps dc->plot #(200 200)))
+                                      #(200 200) 1e-15))
+            (test-suite "PR90: 3d/snip plane-vector"
+                        (check-equal? (send ps plot->dc (send ps plane-vector))
+                                      (send ps plot->dc #(0 0 0))))
+            (test-suite "PR90: 3d/snip before resize"
+                        (check-equal? (send ps plot->dc plotcoords) coords))
+            (test-suite "PR90: 3d/snip after resize"
+                        (send ps resize 800 800)
+                        (sleep/yield .5)
```
alex-hhh, line 73: Unfortunately, this sleep call will create a flaky test which might fail depending on what machine it is run on and the overall CPU load on that machine.  Unfortunately, I don't have a suggestion on how to carry out this test, so I think this should be disabled for now. 

I think the testing is sufficient for now.

bdeket: Will it be ok if I leave the test in as a comment?

alex-hhh: Yes, it should be left in as a comment -- it would be a good idea to be able to test these things, but I don't know how.

Also, you will need to merge the master branch into this one.

samth: There ought to be something that can wait for the resize to complete -- maybe @rfindler knows.

samth:
Returning to the "marking things unstable" issue, after thinking about this more and talking with the rest of the core team, I think that just noting in the documentation that this API is unstable in the 8.1 release is very reasonable. Of course, just committing to it now is also fine, and I'll leave that judgment up to you.

alex-hhh:
I will consider any backwards incompatible changes very carefully, and avoid them if at all possible, but, given that this change will not be validated with real use-cases, there is a small possibility that we got this interface wrong.

dented42:
Sorry for the delay, alas it looks like I missed all the action. Regardless @alex-hhh this handles everything that I was thinking of.

> My concern is that this is a large and complex change, and we only have only one use case for it ([#83 (comment)](https://github.com/racket/plot/issues/83#issuecomment-790017560)). There is a distinct possibility that, apart from perhaps a few tests and toy examples, no one will use this, and, when they will, we'll discover that, despite our best efforts, the interface is not sufficient or not useful.

I'm no expert on API design, but this implementation seems to me to have enough wiggle room that it can be modified or extended with a new approach that feels closer to the platonic ideal of this sort of a thing.

PR #92, opened 2021-04-03
Another attempt to add argument checking for plot-area

sorawee:
Fixes racket/racket#3753
Closes #91

The code is mostly taken from #91.

### DIFF
--- a/plot-lib/plot/private/no-gui/plot2d-utils-typed.rkt
+++ b/plot-lib/plot/private/no-gui/plot2d-utils-typed.rkt
@@ -1,0 +1,93 @@
+#lang typed/racket/base
+
+(require racket/list typed/racket/class racket/match
+         "../common/math.rkt"
+         "../common/plot-element.rkt"
+         "../common/nonrenderer.rkt"
+         "../common/format.rkt"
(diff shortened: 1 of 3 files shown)


alex-hhh:
Closing this PR since it does not address the general problem with TR - Racket interaction and it increases the maintenance burden:

* The general problem is that TR optimizes away checks that are important, see https://github.com/racket/plot/pull/91#issuecomment-812820598
* There are now two files "plot2d-utils.rkt" and "plot2d-utils-typed.rkt" and it is easy to require the wrong one in other code, and neither TR or Racket will flag the problem
* while `plot-area` is protected in "plot2d-utils.rkt", other functions are just re-exported with an `all-from-out` clause, yet they may have the same problems.  In fact, there may be many other functions in other files which have the same problem.
* `plot-area` is an internal function to the plot package and it should not raise exceptions anyway -- the exceptions, just as the crash indicate bugs in the plot package itself and we should focus on fixing the actual bugs.

The actual problem was fixed in this commit: https://github.com/racket/plot/commit/9189c8f71d0585ea46a82f9032e9e706243417eb

PR #93, opened 2021-04-25
Refine hash table's type.

NoahStoryM:
https://github.com/racket/typed-racket/pull/1081


alex-hhh:
It seems these changes are related to the linked pull request in the Typed Racket repository, but I am not sure why those changes are needed.  @samth , I will leave this PR to you.

samth:
I think this PR is fine but not required, since I don't think we're going to do the linked TR change right now. This PR gives more precise types which has advantages, and doesn't break anything.

PR #99, opened 2021-10-08
fix the types fo LVof

capfredf:
change those types to be forward-compatible with TR's upcoming changes where
polymorphic types are not replacements of type constructors any more.


capfredf:
@alex-hhh it should be fine. The change I made is compatible with all past and current version of Typed Racket where `define-type` is supported. And we won't make the linked change set to Type Racket into 8.3. After all, it is still a work-in-progress.

alex-hhh:
OK, I merged it down, but will not add this change to the release branch so it won't be in 8.3

PR #103, opened 2021-11-14
Document the plot-metrics interface

alex-hhh:
Pull request #90 implements the plot-metrics interface, a mechanism that allows the user to determine the location of elements on the plot image itself, so plots can be further decorated, it is based on a request in issue #83.  While this functionality has been available for a while now, it was not documented.    This PR adds documentation to this interface, so potential users can discover the functionality and use it.

## Outstanding issue.

The `plot-metrics<%>` interface itself is not provided by the plot package, and while the functionality is accessible (e.g. calling plot-metrics methods), the user cannot test if a returned image implements this interface (see example below).  Unfortunately, this means that the scribble documentation is incorrectly linked...

```racket
#lang racket
(require plot plot/utils)
(define a-plot (plot-bitmap (function values -1 1)))
(send a-plot get-plot-bounds) ; => '#(#(-1 1) #(-1 1))
(is-a? a-plot plot-metrics<%>) ; => plot-metrics<%>: unbound identifier in: plot-metrics<%>
```

### DIFF
--- a/plot-doc/plot/scribblings/utils.scrbl
+++ b/plot-doc/plot/scribblings/utils.scrbl
@@ -582,3 +582,136 @@ Convert @racket[plot-time]s to real seconds, and vice-versa.
                  (plot-time+ (plot-time 32 0 12 1)
                              (plot-time 32 0 14 1))]
 }
+
+@section{Plot Metrics}
+
+@definterface[plot-metrics<%> ()]{
+
+  The @racket[plot-metrics<%>] interface allows obtaining plot area positions
+  on the plots returned by @racket[plot], @racket[plot-snip],
+  @racket[plot-bitmap], @racket[plot/dc], as well as their 3D variants,
+  @racket[plot3d], @racket[plot3d-snip], @racket[plot3d-bitmap] and
+  @racket[plot3d/dc].  All plot objects returned by these functions implement
+  this interface.
+
+  In addition to this, the plots created by @racket[plot-snip] and
+  @racket[plot3d-snip] have their own set of functions allowing access to the
+  same functionality.
+
+  @defmethod[(get-plot-bounds) (vectorof (vector/c real? real?))]{
+
+    Return the bounds of the plot as a vector of minimum and maximum values,
+    one for each axis in the plot.  For 2D plots, this method returns a vector
+    of two elements, for the X and Y axes, while 3D plots return a vector of
+    three elements for the X, Y and Z axes.
+
+    The values returned are in plot coordinates, to obtain the coordinates on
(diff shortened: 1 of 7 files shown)


plot-lib/plot/private/utils-and-no-gui.rkt
```diff
@@ -21,9 +21,12 @@
  Legend-Anchor)
 
 (require "common/plotmetrics.rkt")
+(require/typed (submod "common/plotmetrics.rkt" untyped)
+  [#:opaque plot-metrics<%> interface?])
```
alex-hhh, line 25: Hi @samth , could you provide some advice on how to export an interface from Typed Racket?   `plot-metrics<%>` is an interface, and I would like to be exported and used from plain Racket when the user requires the `plot` or `plot/no-gui` package.

The code above exports `plot-metrics<%>` but it is a type, and trying to do a `(is-a? o plot-metrics<%>)` fails with "type-check: type name used out of context".   For an example of this failure, see "plot-test/plot/tests/PRs/90.rkt".

Thanks,
Alex.

alex-hhh: This was the first thing I tried, and didn't work.  On further investigation, it seems that the problem is that the line `(provide (all-from-out "private/utils-and-no-gui.rkt"))`  in "utils.rkt" does not actually export the `plot-metrics<%>` interface.  However, adding an explicit provide there fixed the problem.

There are some other problems with the way that interface was defined, I think I will fix them as a separate pull request, to keep this one as a "documentation only" PR.

plot-doc/plot/scribblings/utils.scrbl
```diff
@@ -582,3 +582,136 @@ Convert @racket[plot-time]s to real seconds, and vice-versa.
...
+@definterface[plot-metrics<%> ()]{
+
+  The @racket[plot-metrics<%>] interface allows obtaining plot area positions
+  on the plots returned by @racket[plot], @racket[plot-snip],
+  @racket[plot-bitmap], @racket[plot/dc], as well as their 3D variants,
+  @racket[plot3d], @racket[plot3d-snip], @racket[plot3d-bitmap] and
+  @racket[plot3d/dc].  All plot objects returned by these functions implement
+  this interface.
+
+  In addition to this, the plots created by @racket[plot-snip] and
```
bdeket, line 597: What is this 'own' functionality for plot(3d)-snip?
Or did you mean plot(3d)-pict?

alex-hhh: Yes, I meant `plot-pict` and `plot3d-pict`.  I fixed this and clarified that sentence.  Thanks for picking up the mistake.

PR #104, opened 2021-11-29
remove 'untyped-' redirects

bennn:
Remove extra defines from untyped reprovide modules. These aren't
needed for safety, and I don't think they help much with error
messages & keeping ids distinct either.

(After removing, the `define-typed/untyped-id` in plot work as-is
for Transient. If we kept the defines, transient would need new
identifiers.)

### DIFF
--- a/plot-lib/plot/bitmap.rkt
+++ b/plot-lib/plot/bitmap.rkt
@@ -17,7 +17,7 @@
            "private/plot2d/renderer.rkt")
   (provide
    (contract-out
-    [untyped-plot
+    [plot
      (->* [(treeof (or/c renderer2d? nonrenderer?))]
           [#:x-min (or/c real? #f)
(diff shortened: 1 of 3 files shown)


bennn:
If there's a good reason to keep the `define`s here, let me know. I can open a different PR after https://github.com/racket/typed-racket/pull/948 gets merged to let plot work for the `typed/racket/shallow` lang.

alex-hhh:
Changes look good to me, but I don't know what "Transient" and `typed/racket/shallow` refer to?  I suspect this is related to Typed Racket work?  @samth , if this is OK with you as well, I am happy to merge this.

samth:
Those refer to racket/typed-racket#948, which is @bennn's PR adding new features to Typed Racket.

I've talked to Ben about this change and I think it won't affect anything in Plot.

PR #105, opened 2022-01-07
tests: Update golden test data files to match record-dc% changes

lexi-lambda:
This is necessary since racket/draw#44 added an extra field to `font%` instances, which results in an extra element in the `record-dc%` internal representation.

Since there are so many tests here, I wrote a program to update the `.dat` files conservatively, rather than just blindly accept any new output. More specifically, the program read each `.dat` and updated each `set-font` datum to add `#hash()` at the end of its list and otherwise made no other changes. The tests pass with those minimal modifications, so I am confident there is no actual behavioral change here.

Technically, this PR also makes one tiny other change: I changed the epsilon value in a `check-within` call from `1e-15` to `1e-13`, since `1e-15` is too strict on my machine. With that modification, all the tests pass.


lexi-lambda:
It looks like the 8.2 tests fail because they are unable to install `plot-test` on 8.2, as the 8.2 catalog does not include a new enough version of `draw-lib`. This is unsurprising, seeing as the update to `draw-lib` was made yesterday and has not made it into any release.

I’m not sure how much this is really a problem, since (a) it only affects `plot-test`, not `plot-lib` or `plot`, and (b) my understanding is that `plot-test` is also included in the distribution catalog, anyway. So I’ve updated `.github/workflows/ci.yml` to just not run CI against anything other than HEAD, which seems alright to me (and once 8.4 is released it can be added back to the build matrix), but let me know if you think that isn’t okay and I can figure out something else.

alex-hhh:
Thanks for submitting the PR.   With regards to using a released version in the CI, I wanted to have a stable Racket build against which to test any plot changes, since "current" can sometimes fail, however, this is one of those special cases where the plot package depends on a Racket feature in "current".

I will add back in the stable version once 8.4 is released.

alex-hhh (approved):

PR #106, opened 2022-01-15
Violin plots

camoy:
Here's a first go at implementing violin plots. Let me know what you think, @alex-hhh.


plot-lib/plot/private/plot2d/interval.rkt
```diff
@@ -374,3 +375,89 @@
...
+          #:line2-style Plot-Pen-Style
+          #:alpha Nonnegative-Real
+          #:label (U String pict #f)
+          #:bandwidth (U Real #f)
+          #:invert? Boolean]
+         renderer2d))
+(define (violin
+         x-center ys
+         [y-min #f] [y-max #f]
+         #:x-min [x-min #f] #:x-max [x-max #f]
```
alex-hhh, line 403: I am not sure what the purpose of the `x-min` and `x-max` arguments are.  For a renderer they should control the X range, if it makes sense.   Also, there seems not be a way to control the width of the violin plot (seems to be a width of 1 from -0.5 to 0.5?)

Since the violin plots have essentially one dimension, we need a mechanism to allow the user to reliably place several of them next to each other, similar to how `discrete-histogram` uses  `discrete-histogram-skip` and `discrete-histogram-gap`.  So basically:

* add `#:skip` and `#:gap` parameters to the `violin` plot, they can default to the same `discrete-histogram-skip` and `discrete-histogram-gap`
* the width of the plot should be (- skip gap), so the function interval should be scaled by this amount

alex-hhh (changes requested):
Hi @camoy,thanks for doing this so quickly. Changes look really good,  I have a few comments, aimed to make the `violin` plot easier to use.  There are more details in the comments, but the summary is:

- [x] Update the implementation so that the violin label is added as a tick on the axis
- [ ] compute the y-range automatically, and make `y-min`, `y-max` keyword arguments default to `#f`.  Basically, for a non-empty dataset, `(violin 0 xs)` should just work and display an acceptable plot.
- [ ] remove `x-min` and `x-max` and add a mechanism to control the width of the plot, based on the `skip` + `gap` technique used for `discrete-histogram` plots.

Let me know if something does not make sense (I also make mistakes :-) ), or if you have any questions.

Thanks,
Alex.

alex-hhh:
Now that I think more about it, perhaps the `x` value should also be a keyword argument (perhaps `#:x`), defaulting to 0, so that simply writing `(violin xs)` should get a basic plot up.  What do you think?

plot-lib/plot/private/plot2d/interval.rkt
```diff
@@ -374,3 +375,89 @@
...
+          #:line2-width Nonnegative-Real
+          #:line2-style Plot-Pen-Style
+          #:alpha Nonnegative-Real
+          #:label (U String pict #f)
+          #:bandwidth (U Real #f)
+          #:invert? Boolean]
+         renderer2d))
+(define (violin
+         x-center ys
+         [y-min #f] [y-max #f]
```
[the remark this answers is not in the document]
camoy: It turns out KDE actually returns the bounds, so this will be easy to do.

plot-lib/plot/private/plot2d/interval.rkt
camoy: In general violin plots can have arbitrary width because the KDE can have arbitrary width (the constraint is that the area is 1 but nothing on the maximum width). Taking a look at other plotting libraries it looks like some have an option to do either a fixed width (by scaling) or the full width. So maybe we have a `#:width` that defaults to `(- skip gap)` and if it's explicitly `#f` to use the full width.

As for several violins, do you think it would make more sense to have the data be more in the format of `stacked-histogram`? I.e. instead of just ploting one violin you would do: `(violin (list #("a" (1 2 3)) #("b" (3 4 5)) ...)` and you would get several next to one another. Then `#:skip` and `#:gap` make more sense.

alex-hhh: The value `(- skip gap)` should always represent the width of 1.0 for a violin plot and the plot might extend that width, possibly overlapping with nearby plots.  The requested bounds for the renderer should always be the actual width of the plot (scaled by `(- skip gap)`), so the plot is auto-sized to fit the whole plot by default, with no clipping.  If the user does not want their violin plots to overlap, they can specify larger `x` values between adjacent `violin` plots.  Is there a case where an extra `width` argument is required?

The `skip` and `gap` arguments are most useful when the `violin` plot is used with other renderers and the plot area is not "unit" in size. 

> As for several violins, do you think it would make more sense to have the data be more in the format of stacked-histogram? 

I think it is simpler to just have `violin` plot a single data set, as it does now:

* it is easy to add multiple violin plots and also to use `for/list` to automatically create a multi-violin plot.  
* most novice users will have trouble specifying the data in a correct format, since it would involve using nested lists.  
* for presentation purposes, it is common to use different colors for each violin in a multi-violin plot, so the input data would be even more complex, or users are back to using multiple `violin` renderers.

I think we should keep `violin` as it is now, plotting a single data set.  In the future, we can add a new `violin*` renderer if users request it.

camoy:
Ok, this should be better. Let me know if I did the skip-gap thing correctly, I'm not certain I got it right. If things look good, I'll update the documentation based on these changes.

plot-doc/plot/scribblings/renderer2d.scrbl
```diff
@@ -557,6 +557,38 @@ Corresponds with @(racket polar).
...
+          [#:line1-style line1-style plot-pen-style/c (interval-line1-style)]
+          [#:line2-color line2-color plot-color/c (interval-line2-color)]
+          [#:line2-width line2-width (>=/c 0) (interval-line2-width)]
+          [#:line2-style line2-style plot-pen-style/c (interval-line2-style)]
+          [#:alpha alpha (real-in 0 1) (interval-alpha)]
+          [#:label label (or/c string? pict? #f) #f]
+          [#:invert? invert? boolean? #f]
+          [#:bandwidth bandwidth (or/c real? #f) #f]
+          ) renderer2d?]{
+Draws a violin plot centered at @racket[x].
```
alex-hhh, line 579: When updating this documentation, can you please make the "violin" a link to https://en.wikipedia.org/wiki/Violin_plot

alex-hhh (commented):
Changes look good.  I made a few comments about fewer restrictions on the `gap` parameter and some doc improvements, but other than that, this looks good to me.

plot-lib/plot/private/plot2d/interval.rkt
```diff
@@ -412,30 +419,31 @@
...
   (define fail/kw (make-raise-keyword-error 'violin))
   (cond
-    [(and y-min (not (rational? y-min)))  (fail/pos "#f or rational" 2)]
-    [(and y-max (not (rational? y-max)))  (fail/pos "#f or rational" 3)]
-    [(and x-min (not (rational? x-min)))  (fail/kw "#f or rational" '#:x-min x-min)]
-    [(and x-max (not (rational? x-max)))  (fail/kw "#f or rational" '#:x-max x-max)]
     [(and y-min (not (rational? y-min)))  (fail/kw "#f or rational" '#:y-min y-min)]
     [(and y-max (not (rational? y-max)))  (fail/kw "#f or rational" '#:y-max y-max)]
+    [(or (> gap 1) (not (rational? gap)))  (fail/kw "real in [0,1]" '#:gap gap)]
+    [(not (rational? skip))  (fail/kw "rational?" '#:skip skip)]
```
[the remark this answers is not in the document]
camoy: The type `Nonnegative-Real` should generate a contract that already takes care of the `positive?` restriction.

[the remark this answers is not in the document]
alex-hhh: > The type Nonnegative-Real should generate a contract that already takes care of the positive? restriction

Not always.  I don't think the comment below applies to this file now , but it is safer to add the checks to the entry functions into the plot package, since subtle changes to the code can cause crashes:

https://github.com/racket/plot/blob/6da78b761a6005fdfe771318097513adf38c9564/plot-gui-lib/plot/private/gui/plot2d.rkt#L3

camoy:
Thanks for all your help on this PR! Let me know if there's anything else I should do.

alex-hhh:
Thanks for makin the changes, they are all good.  I will have another more thorough look later this week but I think this is ready to merge.

alex-hhh (approved):

alex-hhh:
Thanks for your contribution, I merged it into the plot package and will be available in the next Racket release (8.5)

PR #115, opened 2023-03-26
doc: defproc for min*, max*

bennn:
these have been available but undocumented for years

- - -

I keep rediscovering them by requiring `plot/utils` and a `min*` / `max*` that expect lists. Adding this entry to the docs will make it easier to learn that `plot/utils` is the exporter.


alex-hhh:
I'll merge this, since it is just a documentation update, but perhaps `min*` and `max*` should be outside of the plot package, defined together with `min` and `max`

PR #123, opened 2023-06-30
Allow controlling line ending caps for plots (#6)

alex-hhh:
New `plot-cap` and `plot-line-cap` control the line endings for lines in the plot.  Previously all line endings were `'round`:

6-round

Now they can be chosen by the user (one for the plot decorations, such as axes and ticks, another for the lines on the plot):

```racket
#lang racket
(require plot)
(plot-width 800)
(plot-height 600)

(parameterize
    ([plot-line-width 15]
     [line-width 15]
     [plot-pen-color-map 'set1]
     [plot-line-cap 'projecting]
     [line-cap 'butt])
  (plot
   (list
    (function sin -5 5 #:color 0 #:label "sin(x)")
    (function cos -5 5 #:color 1 #:label "cos(x)"))))
```


alex-hhh:
This also fixes a small bug with `vrule` and `hrule` renderers, which attempted to override the line endins for their lines to `'butt` insead of `'round`: since plot pens are kept in a hash table, which is not indexed by line cap, if a pen with the same width and color was created before a `vrule` renderer, the `vrule` would have round endings, if `vrule` created the pen, other renderers which used the same width and color would have `'butt` line endings.  This is now fixed and `vrule`/`hrule` renderers use the value of `line-cap`

alex-hhh:
A noticeable effect when using very thick lines is that the line edges are clipped for both the plot area and the legend.  This is visible in both screenshots above.  However, this is a different problem not linked to these changes and a fix for that will be done as a separate PR.

Issue #126, opened 2024-01-26
Differences in text size


sorawee:
When I run tests under https://github.com/racket/plot/tree/master/plot-test/plot/tests and force regeneration, the new images have very small text size compared to the old images, even though all tests passed.

With Racket 8.11.1 on Mac:

Newly generated `new-p3d-bsp-9`:
![new-p3d-bsp-9](https://github.com/racket/plot/assets/9099577/45a42c1c-6674-41bf-be5b-f8e18644dab5)

Original `p3d-bsp-9` (from https://github.com/racket/plot/tree/master/plot-test/plot/tests/test-data):
![p3d-bsp-9](https://github.com/racket/plot/assets/9099577/d44f4087-0df5-4d79-9b0d-81783c87d787)

alex-hhh:
This is expected behavior, even if it is surprising:  default fonts, and font rendering, is different on different platforms.  Usually plot images are generated on a linux VM (Where the github action runs) or on my Windows machine.

The generated plot steps ignore font rendering, by faking character with and heights, so the resulting draw steps themselves are identical on different platforms.   The images themselves are only used (usually by me) as a quick visual check if a test fails, and I know not to look at the font size :-).

https://github.com/racket/plot/blob/2417c228b0ce4708c796b4cb0a22cc818932a605/plot-test/plot/tests/helpers.rkt#L23

sorawee:
Got it!

PR #127, opened 2024-01-27
test: add a way to regenerate test data

sorawee:


alex-hhh:
Not sure what you aim with this PR, but I wanted it to make it more difficult to quickly override images and draw steps for failed tests.

Currently, the developer has to stop and think and manually verify that each of the failing tests don't introduce actual visual problems with the plot.  With your proposed change it becomes too easy for a developer to simply decide "I'm sure my changes are OK, let me just update all these tests at once and go..."

sorawee:
My two cents: 

They won't be able to "quickly override images and draw steps for failed tests", because there is a review process.

On the other hand, overriding the file allows one to compare images in GitHub diff view, which is much easier than finding two specific files from a directory with hundred files, and then arrange windows to compare them side by side.

To me, this is simply a convenience tool.

alex-hhh:
I considered adding this feature when I originally implemented these tests but decided not to.  

Making it too convenient to update these tests introduces the risk of putting backwards incompatible changes in the plot package, where the output of the plot changes between releases (even if the API stays the same).

The current setup is not ideal, but it does not seem to prevent or hinder people from contributing features and fixes to the plot package (many contributions were done since I introduced this system several years ago).   If this becomes a practical problem in the future, I'll reconsider it.

---

> They won't be able to "quickly override images and draw steps for failed tests", because there is a review process.

The review process is not perfect and can fail too.  In my professional field, the systems we design have multi-layered safety features and one cannot justify removing safety features just because "there is at least one safety feature remaining."  

> On the other hand, overriding the file allows one to compare images in GitHub diff view, which is much easier than finding two specific files from a directory with hundred files, and then arrange windows to compare them side by side.

Comparing images is easy in Github, but as you noted yourself in #126 , images will never be identical, and it is easy to look at images, and decide that any changes on the image are caused by platform specific differences. 

The ".DAT" files, which contain the rendering steps, are the true indicator of what changed, and if a change is backwards compatible or not.  Unfortunately, these binary files cannot be diff-ed in Github, since they are compressed, to save in file size.

Making it "convenient" would make people rely on the Github image diff, which is the wrong thing to do.

sorawee:
Thanks for the explanation!

PR #128, opened 2024-01-27
Remove unused code in glyph drawing code

alex-hhh:
The original intent of the `when` block was to remove the line around the glyph when the line and fill color were the same. However, the `eq?` test was always false in Typed Racket, so the code never executed.

The `eq?` test is now fixed in
https://github.com/racket/typed-racket/pull/1358, but as a result, the visual layout of plots would change, making glyphs smaller, since the outline is not drawn anymore.

To keep the backwards compatibility of the plot look-and-feel, the entire `when` block was removed.

### DIFF
--- a/plot-lib/plot/private/common/plot-device.rkt
+++ b/plot-lib/plot/private/common/plot-device.rkt
@@ -546,9 +546,6 @@
              (define r (* 1/2 size))
              (define line-sym
                (cond [(hash-has-key? full-glyph-hash real-sym)
-                      (when (eq? pen-color brush-color)
-                        (set-pen pen-color 1 'transparent)
-                        (set-brush brush-color 'solid))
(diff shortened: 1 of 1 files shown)


alex-hhh:
@sorawee FYI

sorawee:
Interestingly, DrDr has still been failing, even after this PR. 

http://drdr.racket-lang.org/65790/

@samth do you know what's going on?

alex-hhh:
@sorawee is there a way in DrDr to get access to the build artifacts?  I downloaded the archive (trunk.tgz), but that seems to only contain the source code being built...  It would be good to compare the draw steps on the built to determine what failed.

samth:
I believe that it was just the pkg server not updating; it looks fixed now.

PR #129, opened 2024-03-29
Repair errors uncovered by racket/typed-racket#1367.

samth:
Some arguments were not type checked correctly, meaning some type
errors slipped through.

The most significant change is that the jitter parameters become
`Nonnegative-Real`, which is backwards-incompatible. Other choices
there are possible, but this is the simplest.


plot-lib/plot/private/plot3d/point.rkt
```diff
@@ -51,7 +51,7 @@
                   #:sym [sym (point-sym)]
                   #:color [color (point-color)]
                   #:fill-color [fill-color 'auto]
-                  #:x-jitter [x-jitter (point-x-jitter)]
+                  #:x-jitter [#{x-jitter : Nonnegative-Real} (point-x-jitter)]
```
alex-hhh: Should the `Nonnegative-Real` type be applied to the `y-jitter` and `z-jitter` arguments below?

alex-hhh (approved):

plot-lib/plot/private/plot3d/point.rkt
samth: That type annotation was unnecessary after I fixed other things; I've removed it.

